PR #20, opened 2015-02-19
Change trigger_job to accept auth as parameter

nigelbabu:
Currently, using the script results in a python error.


armenzg:
Thanks for your patch, however, I landed this fix this morning.
Could you please let me know if it fixes your issue?
https://github.com/armenzg/mozilla_ci_tools/commit/6c3aa8a1c3e546457ad0e3d2de9dc8ecbc59e2b8

PR #24, opened 2015-02-20
Adding comments

adusca:


armenzg:
adusca, the Travis CI information for this pull requests shows that we're failling.
Could you please have a look at that?
Could you also merge changes from master? The pull request says "We can’t automatically merge this pull request."

mozci/platforms.py
```diff
@@ -9,26 +9,69 @@
 
 LOG = logging.getLogger()
 
+# We will start pre-computing some structures to make
+# associated_build_job faster.
```
armenzg: More than faster, we're trying to provide more accurate sources.

mozci/platforms.py
```diff
@@ -9,26 +9,69 @@
...
+
 data = sources.allthethings._fetch_json()
```
armenzg: No need to fetch the data since list_builders() already does that internally:
https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/sources/allthethings.py#L55

armenzg: I see why you do this.
Let's change _fetch_json() to something not starting with "_" since that implies private function (implicit meaning).
Maybe "fetch_allthethings_data".

armenzg: We could then change the variable "data" to be "all_builders_information".

mozci/platforms.py
```diff
@@ -9,26 +9,69 @@
...
 builders = sources.allthethings.list_builders()
```
armenzg: Could you change the import line number 6 to say:

> from mozci.sources.allthethings import fetch_allthethings_data
> Over here we would call builders = all_builders_information.keys()

If it complains about relative paths, I have added this to some files:

> from **future** import absolute_import

mozci/platforms.py
```diff
@@ -9,26 +9,69 @@
...
+# job triggers is not directly available from the json file. Since
+# trigger names for a given build are similar to their shortnames,
+# which are available in allthethings.json, we'll use shortnames to
+# find the which builder triggered a given scheduler given only the
+# trigger name. In order to do that we'll need a mapping from
+# shortnames to build jobs. For example:
+# "Android armv7 API 11+ larch build":
+#               { ...  "shortname": "larch-android-api-11", ...},
+# Will give us the entry:
+# "larch-android-api-11" : "Android armv7 API 11+ larch build"
```
armenzg, line 31: Beautifully explained :+1:

mozci/platforms.py
```diff
@@ -9,26 +9,69 @@
...
 
+# We'll look at every builder and if it's a build job we will add it
+# to shortname_to_name
 for build in builders:
-    # Skipping these for now
+    # Skipping nightly for now
     if 'nightly' in build:
         continue
 
     builder = data['builders'][build]
```
armenzg: We have in this block the variables "build", "builders" and "builder".

Maybe call "builder" as "builder_info".
Call "build" as "buildername" (which generally implies a string for people used to buildbot).
Maybe "builders" to be called "buildernames" (which we can think of a list of strings).

test/test_platforms.py
```diff
@@ -19,16 +19,31 @@ def complex_data():
     with open(filepath, 'r') as f:
         reference_builders_info = json.load(f)
 
-    # Build jobs should return themselves
+    builders_data = mozci.sources.allthethings._fetch_json()['builders']
```
armenzg: What I mentioned above of using a private method.

armenzg:
@adusca I'm done with my comments.
Please let me know when you have addressed and remember to run "tox" locally to make sure that TravisCI will come back successfully.

adusca:
@armenzg Thanks for the comments! I think I addressed them all.

armenzg:
Awesome!
Thank you so much. Let me know if there is anything else that catches your attention.
There might be other issues that would be interesting for you.

PR #64, opened 2015-02-27
Created new test mocking allthethings.py to run with travis

adusca:
Our current test has to download allthethings.json, and that doesn't work well with TravisCI. I created a smaller test (that mocks allthethings.json) inside test/travis/ and changed tox configuration to just run py.test inside that folder.


adusca:
One problem is that the mocking works only if the import statement comes after the mock function, and pep8 does not like imports in the middle of the file.

vaibhavmagarwal:
@adusca, see my comment in https://github.com/armenzg/mozilla_ci_tools/issues/59#issuecomment-76344567 , I think we will actually want travisCI to download the file and run the tests rather than have a mock test file. The reason is, if allthethings.json changes, we need to edit the mock-test everytime. Also, the problem that travis has can also be faced by a developer locally.
Maybe we can use this test as a unit-test.

armenzg:
@adusca: Let's try what @vaibhavmagarwal is trying and see if it works.

@vaibhavmagarwal We will need a pull request reactivating the unit tests.
Let's first _only_ request a PR for enabling py.test. We should see the TravisCI and hopefully it is good.

After that we can have a PR for enabling coveralls. I want to see that we fix one thing at a time.

Thanks to both of you!

adusca:
I also prefer @vaibhavmagarwal 's approach, it will make things more robust.

PR #67, opened 2015-02-27
The string representation of a UTC time includes UTC in it

valeriat:
This pull request fixes issue#44.

### DIFF
--- a/mozci/utils/tzone.py
+++ b/mozci/utils/tzone.py
@@ -62,7 +62,16 @@ class Pacific(USTimeZone):
         USTimeZone.__init__(self, -8, "Pacific",  "PST", "PDT")
 
 
+class Universal(USTimeZone):
+    def __init__(self):
+        USTimeZone.__init__(self, -0, "Universal",  "UTC", "UTC")
+
+    def dst(self, dt):
+        return ZERO
+
+
 pacific_tz = Pacific()
+universal_tz = Universal()
 time_format = '%a, %d %b %Y %H:%M:%S %z (%Z)'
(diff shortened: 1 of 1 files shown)


vaibhavmagarwal:
@valeriat : your patch seems to be not rebased on the latest code. Can you please rebase it and send in the pull request. Thanks

mozci/utils/tzone.py
```diff
@@ -62,7 +62,16 @@ def __init__(self):
         USTimeZone.__init__(self, -8, "Pacific",  "PST", "PDT")
 
 
+class Universal(USTimeZone):
```
armenzg, line 65: Can you please inherit from datetime.tzinfo?

armenzg:
I'm glad to see you made progress in here. Thank you!
A `git pull` is needed as @vaibhavmagarwal pointed out.

mozci/utils/tzone.py
armenzg, line 65: Perhaps USTimeZone should be transformed into a more generic class. I don't know.
Perhaps there are some libraries out there from which we can copy.
Or drop the use of tzone.py and include one of those libraries.
I don't want to scope creep.

vaibhavmagarwal:
Hey @valeriat, let us know if you face any problems dealing with this, and would like our help on this.
You would need to do:
git checkout -b UTC-problem
git rebase master

It will possibly have conflicts that will needed to be solved manually. We normally are present in #ateam on irc.mozilla.org, so drop by if you have questions!

valeriat:
Thank you! I made new pull request.

PR #71, opened 2015-02-27
Added missing return statement to _fetch

adusca:
_fetch was returning None when _verify_file_integrity was False, because we weren't propagating the result of retrying.


vaibhavmagarwal:
Thanks @adusca , good observation!

Issue #44, opened 2015-02-24
The string representation of a UTC time should include UTC in it


armenzg:
Currently:
>>> from mozci.utils.tzone import utc_time, pacific_time
>>> pacific_time()
'Tue, 24 Feb 2015 08:22:01 -0800 (PST)'
>>> utc_time()
'Tue, 24 Feb 2015 16:23:22  ()'

What I expect:
>>> utc_time()
'Tue, 24 Feb 2015 16:23:22 -0000 (UTC)'

valeriat:
Thanks! I would like to try.

vaibhavmagarwal:
Hi @valeriat, thanks for your interest in this bug!

To solve this bug, you will need to do 2 things:

1) Make utc_dt function[1] similar to pacific_time function[2]. So define universal_tz instead of pacific_tz.

2) Define Universal class like Pacific class[3].

[1] - https://github.com/armenzg/mozilla_ci_tools/blob/fa4ca3c536746a6168e1968c0325965404a185ca/mozci/utils/tzone.py#L92
[2] - https://github.com/armenzg/mozilla_ci_tools/blob/fa4ca3c536746a6168e1968c0325965404a185ca/mozci/utils/tzone.py#L70
[3] - https://github.com/armenzg/mozilla_ci_tools/blob/fa4ca3c536746a6168e1968c0325965404a185ca/mozci/utils/tzone.py#L60

This should solve the bug. To test, just call pacific_time() and utc_time() function in any script inside scripts/ folder. Let me know how it goes and if you need any help!

valeriat:
Thank you, vaibhavmagarwal! 
There is one more problem with PST. We calculate time and then we determine format, so there is transition from DST to standart timet at 01:00:00 PDT not PST.

For example: 
Sun, 25 Oct 2015 00:59:59 -0700 (PDT) then Sun, 25 Oct 2015 01:00:00 -0800 (PST)
Sun, 25 Oct 2015 01:59:59 -0800 (PST) then Sun, 25 Oct 2015 01:00:00 -0800 (PST) 
Time offset is correct but PDT - PST transition is late for an hour.

vaibhavmagarwal:
I see. There maybe some error in logic in the pdt and pst conversion. @valeriat can you look into tzone.py and figure out what is going wrong?
@armenzg do you see the same issue and can you help here?

armenzg:
Is this how it should work?
[...]

PR #77, opened 2015-03-03
The string representation of a UTC time includes UTC in it #2

valeriat:
This pull request fixes issue #44.

### DIFF
--- a/mozci/utils/tzone.py
+++ b/mozci/utils/tzone.py
@@ -62,7 +62,30 @@ class Pacific(USTimeZone):
         USTimeZone.__init__(self, -8, "Pacific",  "PST", "PDT")
 
 
+class Universal(datetime.tzinfo):
+    def __init__(self):
+        hours = -0
+        reprname = "Universal"
+        stdname = "UTC"
+        self.stdoffset = datetime.timedelta(hours=hours)
+        self.reprname = reprname
+        self.stdname = stdname
+
+    def __repr__(self):
+        return self.reprname
+
+    def tzname(self, dt):
+        return self.stdname
+
+    def utcoffset(self, dt):
+        return self.stdoffset + self.dst(dt)
+
+    def dst(self, dt):
+        return ZERO
+
+
 pacific_tz = Pacific()
+universal_tz = Universal()
 time_format = '%a, %d %b %Y %H:%M:%S %z (%Z)'
 day_format = '%Y-%m-%d'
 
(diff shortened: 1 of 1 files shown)


mozci/utils/tzone.py
```diff
@@ -91,15 +114,17 @@ def pacific_day(timestamp=None):
 
 def utc_dt(timestamp=None):
     if not timestamp:
-        return datetime.datetime.utcnow()
-    else:
-        return datetime.datetime.utcfromtimestamp(timestamp)
+        timestamp = time.time()
```
armenzg, line 117: Please use datetime.datetime.utcnow() instead of time.time().
The former is UTC time and the latter is local time.

mozci/utils/tzone.py
```diff
@@ -91,15 +114,17 @@ def pacific_day(timestamp=None):
...
+    dt = datetime.datetime.fromtimestamp(timestamp, universal_tz)
+
+    return dt
 
 
 def utc_time(timestamp=None):
     """Convert a time expressed in seconds since the epoch to a string representing UTC time.
     If timestamp is not provided or None, the current time as returned by time() is used.
     """
+
```
armenzg: Please remove these extra blank lines.

armenzg:
Great to see this! I will merge it once address the two comments I made.

Thanks Valeria!

vaibhavmagarwal:
Thanks @valeriat for the pull request! You can edit the comments made my armenzg and do:
git commit -a --amend 
to append the changes to the same commit. Then a 'git push' will make changes in this pr itself, automatically.

mozci/utils/tzone.py
valeriat, line 117: @armenzg, are you sure? According to documentation time.time() "returns the time in seconds since the epoch as a floating point number." Which is time zone independent.

valeriat:
@vaibhavmagarwal thank you for your help with git!

mozci/utils/tzone.py
vaibhavmagarwal, line 117: I think @valeriat is correct, since the value returned will always be the same on every machine at the same time with time.time()
And since we are converting it to utc date via: dt = datetime.datetime.fromtimestamp(timestamp, universal_tz)
the date and time will be according to UTC.

refer: http://stackoverflow.com/questions/11845803/is-pythons-time-time-timezone-specific

armenzg, line 117: Thank you both for not taking what I said as the truth!
My apologies. Let's take this.

vaibhavmagarwal:
Wohoo! @valeriat thanks for sticking and closing this! If you would like, there is a new cleanup bug you can work on: https://github.com/armenzg/mozilla_ci_tools/issues/76

valeriat:
Glad to help! :)
@vaibhavmagarwal can you take a look at my pull request #80? :)

vaibhavmagarwal:
Cool. I am taking a leave for today, will review it tomorrow morning!

Issue #76, opened 2015-03-03
Improve how we determine what to trigger


armenzg:
We currently have an if/block to deal differently build jobs (upstream jobs) from other jobs (downstream jobs).
https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/mozci.py#L269

We should simply be able to call without the if/block
            trigger, files = _determine_trigger_objective(
                repo_name,
                revision,
                buildername,
            )

armenzg:
I was working on this code base and I fixed it:
https://github.com/armenzg/mozilla_ci_tools/commit/d538e466ee4af69f8f7b7d09657efb41e27af9b0

PR #80, opened 2015-03-03
Improve how we determine what to trigger

valeriat:
This pull request fixes issue #76

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -11,7 +11,7 @@ from __future__ import absolute_import
 import json
 import logging
 
-from mozci.platforms import determine_upstream_builder, is_downstream
+from mozci.platforms import determine_upstream_builder
 from mozci.sources import allthethings, buildapi, buildjson, pushlog
 from mozci.utils.misc import _all_urls_reachable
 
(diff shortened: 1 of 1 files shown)


valeriat:
How can I test this change? I already run tox.

vaibhavmagarwal:
@valeria, I am not very familiar with how the build jobs and other jobs get triggered, I will let @armenzg review this patch. For basic testing, you can see whether scripts/trigger.py runs properly.
Eg:
python scripts/trigger.py --buildername "b2g_ubuntu64_vm cedar debug test gaia-js-integration-5" --dry-run --revision d2d9810cf8d7

armenzg:
If she does not have LDAP credentials it becomes hard to test it.
I will review it and test it by the end of today.

armenzg:
From looking at the code, I was assuming that we would need changes in _determine_trigger_objective(). It would not be possible for you to test this without LDAP credentials.
I wish we had a way to setup buildapi locally and have sample data.
Long long time ago I tried it and I had some of the DB dump in here:
http://people.mozilla.org/~armenzg/permanent/databases
but setting up buildapi is not something I managed to make it work.

adusca:
Since determine_upstream_builder() of a build job returns itself, when we pass a build job to _determine_trigger_objective() it returns buildername, None.
I think we had the if condition because of the old associated_build_job.

armenzg:
@valeriat ignore the comments about buildapi.
For the rest, did it make sense the remaining comments?

armenzg:
Hi @valeriat 
Thanks for your help, however, I have fixed this code while I was doing some refactoring:
https://github.com/armenzg/mozilla_ci_tools/commit/d538e466ee4af69f8f7b7d09657efb41e27af9b0

Please have a look at https://github.com/armenzg/mozilla_ci_tools/issues for other issues to help with.

PR #85, opened 2015-03-04
Issue 82: Make determine_upstream_builder case-insensitive.

adusca:


armenzg:
I'm concerned about this change.
I would be surprised if we could actually trigger a job on buildapi/self-serve by posting a buildername that is not a valid buildername.

Could we take the change without lowering the buildername? I'm going to make a comment on the issue.

adusca:
With this PR determine_upstream_build never returns an invalid buildername. It just uses .lower() for comparison and in buildername_to_trigger, but the return values come from shortname_to_name or build_jobs, that keep the original capitalization:
        shortname_to_name[builder_info['shortname']] = buildername
        build_jobs.append(buildername)

armenzg:
This looks good. Thanks adusca.

PR #88, opened 2015-03-05
Issue 69: Build graph of platform-to-tests relations

adusca:
This script generates the following json: https://pastebin.mozilla.org/8824385
It's not as stratified as the example @armenzg gave, but it's a start.


armenzg:
Wow! This is not as stratified but it is a very good start.
I think I would put this file into scripts/misc/ for now rather than make it a mozci module.
Some of these functions are going to be an excellent addition to platforms.py!

Thank you Alice. This is going to make a great difference for a lot of people this year. Next year we might be moving to another system instead of Buildbot but for now it will be golden!

mozci/graph.py
```diff
@@ -0,0 +1,40 @@
+import collections
+import json
+
+from sources.allthethings import fetch_allthethings_data
+
+
+def get_test(buildername):
```
armenzg: Add an "_" at the beginning of this function since it is not one of the main ones.

mozci/graph.py
```diff
@@ -0,0 +1,40 @@
...
+    '''For test jobs, the test type is the last part of the name. For example:
+    in Windows 7 32-bit mozilla-central pgo talos chromez-e10s
+    the test type is chromez-e10s
+    '''
+    return buildername.split(" ")[-1]
+
+
+def build_tests_per_platform_graph():
```
armenzg: We need a docstring in here.

Could you please move this function into platforms.py?

mozci/graph.py
```diff
@@ -0,0 +1,40 @@
...
+    builders = fetch_allthethings_data()['builders']
+    graph = collections.defaultdict(list)
+
+    for builder in builders:
+        props = builders[builder]['properties']
```
armenzg: I think you can just do `builder["properties"]` and it would do what you want.

mozci/graph.py
```diff
@@ -0,0 +1,40 @@
...
+        if 'slavebuilddir' in props and props['slavebuilddir'] == 'test':
```
armenzg: Can we use "is_downstream()" from platforms.py?
Modify is_downstream() if need be.

mozci/graph.py
```diff
@@ -0,0 +1,40 @@
...
+            platform = builders[builder]['properties']['platform']
+
+            # For talos tests we have to check stage_platform instead
+            if 'talos' in builder:
+                platform = builders[builder]['properties']['stage_platform']
+
+            test = get_test(builder)
+            if test not in graph[platform]:
+                graph[platform].append(get_test(builder))
```
armenzg: Could you move this block's functionality into platforms.py?
Please create a function called "get_associated_platform_name(buildername)".
That function should work for both build and test jobs.

mozci/graph.py
```diff
@@ -0,0 +1,40 @@
...
+
+    for platform in graph:
+        graph[platform].sort()
+
+    return graph
+
+
+with open('graph.json', 'w') as f:
+    graph = build_tests_per_platform_graph()
+    json.dump(graph, f, sort_keys=True, indent=4, separators=(',', ': '))
```
armenzg: Scripts generally have this:
if **name** == '**main**':
    main()

This allows for a script to be both runnable as well as importable as a module.

adusca:
I think I addressed every comment.

armenzg:
Thanks Alice. I will look at this tomorrow.

mozci/platforms.py
```diff
@@ -122,15 +130,36 @@ def determine_upstream_builder(buildername):
     raise Exception("No build job found.")
 
 
-def is_downstream(buildername):
-    ''' Determine if a job requires files to be triggered.
+def get_associated_platform_name(buildername):
+    '''Given a buildername, find the platform in which it is ran'''
```
armenzg, line 135: Does this also work with build jobs?

armenzg, line 135: I don't see you using it below for build jobs so let's add a note saying that we don't support build jobs (Unless we actually do :D)

armenzg:
Good to see those tests!
Once I get an answer to my questions feel free to merge!

adusca:
It does support build jobs. I added a test case for that.

PR #92, opened 2015-03-05
Remove disposable branches

adusca:


armenzg:
Thank you! This should also speed up our tox runs.

PR #95, opened 2015-03-06
Remove unused repo_name from test_platforms.py

adusca:


armenzg:
Thanks for the cleanup!

PR #103, opened 2015-03-06
Issue 101: Wrap code inside _process_data()

adusca:
I moved computing the dictionaries to the function _process_data(). Also, instead of using all_builders_information we are now just calling fetch_allthethings_data(). To avoid making things too slow I added some in-memory-caching to allthethings.py. The final result is actually faster then before this patch (py.test went from 16s to 7s on my machine)

### DIFF
--- a/mozci/platforms.py
+++ b/mozci/platforms.py
@@ -6,17 +6,11 @@ import logging
 from sources.allthethings import fetch_allthethings_data
 
 LOG = logging.getLogger()
-# We will start by pre-computing some structures that will be used for
-# determine_upstream_builder. They are globals so we don't compute them over
-# and over again when calling determine_upstream_builder multiple times
-
-all_builders_information = fetch_allthethings_data()
-buildernames = all_builders_information['builders'].keys()
 
 
 def is_downstream(buildername):
     """Determine if a job requires files to be triggered."""
-    props = all_builders_information['builders'][buildername]['properties']
+    props = fetch_allthethings_data()['builders'][buildername]['properties']
     return 'slavebuilddir' in props and props['slavebuilddir'] == 'test'
 
 # In buildbot, once a build job finishes, it triggers a scheduler,
@@ -32,7 +26,7 @@ def is_downstream(buildername):
 #               { ...  "shortname": "larch-android-api-11", ...},
 # Will give us the entry:
 # "larch-android-api-11" : "Android armv7 API 11+ larch build"
-shortname_to_name = {}
+SHORTNAME_TO_NAME = {}
 
 # For every test job we can find the scheduler that runs it and the
 # corresponding trigger in allthethings.json. For example:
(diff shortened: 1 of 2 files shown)


mozci/platforms.py
```diff
@@ -51,44 +45,47 @@ def is_downstream(buildername):
 # every test job. In this case, "Android 4.0 armv7 API 11+ larch opt
 # test cppunit" : larch-android-api-11-opt-unittest
 buildername_to_trigger = {}
```
armenzg: Could you please upper case this variable? This is to meet higher pep8 score.

mozci/platforms.py
```diff
@@ -51,44 +45,47 @@ def is_downstream(buildername):
...
-# trigger name in triggered_by
-for sched, values in all_builders_information['schedulers'].iteritems():
-    # We are only interested in test schedulers
-    if not sched.startswith('tests-'):
-        continue
-
-    for buildername in values['downstream']:
-        assert buildername not in buildername_to_trigger
-        buildername_to_trigger[buildername.lower()] = values['triggered_by'][0]
+build_jobs = {}
```
armenzg: Could you please upper case this variable? This is to meet higher pep8 score.

mozci/platforms.py
```diff
@@ -51,44 +45,47 @@ def is_downstream(buildername):
...
+
+
+def _process_data():
+    '''Filling the dictionaries used by determine_upstream_builder. '''
+    # We check if we already computed before
+    if buildername_to_trigger:
+        return
```
armenzg, line 52: Could you please add a debug message?

mozci/platforms.py
```diff
@@ -51,44 +45,47 @@ def is_downstream(buildername):
...
+
+    # We'll look at every builder and if it's a build job we will add it
+    # to shortname_to_name
+    for buildername, builderinfo in fetch_allthethings_data()['builders'].iteritems():
```
armenzg, line 57: Could you please add a debug message that we're running this?

mozci/sources/allthethings.py
```diff
@@ -58,6 +58,8 @@
 ALLTHETHINGS = \
     "https://secure.pub.build.mozilla.org/builddata/reports/allthethings.json"
 
+_data = {}
```
armenzg: Could you please upper case this variable? This is to meet higher pep8 score.

mozci/sources/allthethings.py
```diff
@@ -93,21 +95,25 @@ def _verify_file_integrity():
         else:
             return True
 
+    global _data
     if no_caching:
-        data = _fetch()
```
armenzg, line 97: I realized that we can improve the readability of this block by doing this:

> if no_caching or not os.path.exists(FILENAME)
> since both cases do _data = _fetch()

mozci/sources/allthethings.py
```diff
@@ -93,21 +95,25 @@ def _verify_file_integrity():
...
             # If corrupt incomplete / old file, remove
             if not _verify_file_integrity():
                 os.remove(FILENAME)
-                data = _fetch()
+                _data = _fetch()
             else:
                 fd = open(FILENAME)
-                data = json.load(fd)
+                _data = json.load(fd)
         else:
```
armenzg, line 112: If we change the if/condition we can remove this "else" clause.

armenzg:
Alice, these are excellent news! I'm happy to hear that it makes such a big difference.
I should have caught this in our very first initial reviews. My apologies.

Feel free to merge once you address my comment. Obrigado!

adusca:
I addressed your comments, but I ended up changing more things in the fetch logic. I want a second opinion on these changes, so I'm not merging it immediately (but no hurry!). The reason I didn't merge the two _fetch cases is that I wanted to avoid touching the file whenever possible to make things quicker.

mozci/sources/allthethings.py
```diff
@@ -84,6 +87,9 @@ def _fetch():
             return _fetch()
 
     def _verify_file_integrity():
+        if not os.path.exists(FILENAME):
+            return False
+
```
armenzg, line 92: Don't do this. See in a comment below why.

mozci/sources/allthethings.py
```diff
@@ -93,21 +99,20 @@ def _verify_file_integrity():
...
-            # If corrupt incomplete / old file, remove
-            if not _verify_file_integrity():
-                os.remove(FILENAME)
-                data = _fetch()
-            else:
-                fd = open(FILENAME)
-                data = json.load(fd)
+        DATA = _fetch()
+    # If we do not have an in-memory cache, try to use the file cache.
+    elif not DATA:
```
armenzg: A more pythonic way of writing this is "elif DATA is not None".

Could you please fix that?

> > > a = {}
> > > print not a
> > > True
> > > print a is not None
> > > True
> > > print a == {}
> > > True

mozci/sources/allthethings.py
```diff
@@ -93,21 +99,20 @@ def _verify_file_integrity():
...
+        # Only use the file cache if it is up-to-date and not corrupted.
+        if _verify_file_integrity():
+            fd = open(FILENAME)
+            DATA = json.load(fd)
         else:
-            data = _fetch()
+            DATA = _fetch()
```
armenzg, line 113: We need to delete the file if the verification is incorrect.
We're loosing that with this new approach.

adusca:
Done!

PR #105, opened 2015-03-08
Issue 94: Caching repositories.txt in memory

adusca:
The variable 'repositories' starts as an empty dictionary that gets filled when we run query_repositories(). After that, if clobber=False, the function will return the value stored in the dictionary instead of reading from the file. Now "Loading /home/alice/mozilla_ci_tools/repositories.txt" only shows up once in debug mode.

### DIFF
--- a/mozci/sources/buildapi.py
+++ b/mozci/sources/buildapi.py
@@ -24,6 +24,7 @@ from mozci.sources.buildjson import query_job_data
 LOG = logging.getLogger()
 HOST_ROOT = 'https://secure.pub.build.mozilla.org/buildapi/self-serve'
 REPOSITORIES_FILE = os.path.abspath("repositories.txt")
+REPOSITORIES = {}
 
 # Self-serve cannot give us the whole granularity of states; Use buildjson where necessary.
 # http://hg.mozilla.org/build/buildbot/file/0e02f6f310b4/master/buildbot/status/builder.py#l25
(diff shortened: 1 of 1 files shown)


vaibhavmagarwal:
Shouldn't we keep repo_name as a parameter to trigger_range in the scripts [1] rather than having it in mozci [2]? Since the repo_name remains same for a given buildername, it seems to me like an unnecessary call inside trigger_range function in triggering every job. Then, I think this patch optimization won't be required and we can escape having to make "repositories" as a global variable, as having global variables is generally not good from a code maintenance point of view.

[1] https://github.com/armenzg/mozilla_ci_tools/blob/master/scripts/trigger_range.py#L160
[2] https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/mozci.py#L360

mozci/sources/buildapi.py
```diff
@@ -24,6 +24,7 @@
 LOG = logging.getLogger()
 HOST_ROOT = 'https://secure.pub.build.mozilla.org/buildapi/self-serve'
 REPOSITORIES_FILE = os.path.abspath("repositories.txt")
+repositories = {}
```
armenzg: Uppercase please.

armenzg:
I understand your concerns, however, it is fine as long as it is only use within one module.
In other words, it is generally not a good practice since a lot of people will end up using unapproprietately (reading/changing a global variable all over the place - or using globals frequently for various things).

It would be bad if we started using repositories directly from other modules. For other modules, it is unimportant how you implement it internally.

armenzg:
I don't want scripts having to calculate "repo_name" externally before calling trigger_range().
The less boilerplate, the easier the modules will be to use by new people.

I hope it makes sense; feel free to counter.

armenzg:
If it helps, I spoke with wlach specifically on how to solve this and if it would be a bad practice or not.

vaibhavmagarwal:
@armenzg, cool. I just had some concerns which I wanted to be clarified, it is ok if we go forward with this approach.

armenzg:
Thank you @vaibhavmagarwal for rasing them!

adusca:
Rebased, upper-cased it and fixed a logic mistake.

I never know when it's OK to use global variables. It's good to read these discussions, I learned a lot this way.

armenzg:
Excellent! Many thanks!

PR #106, opened 2015-03-08
Fixed some flake8-docstrings warnings.

adusca:
With this patch, "flake8 mozci scripts test" output is: https://pastebin.mozilla.org/8824832 . There are still 58 errors.

According to pep257, multi-line docstrings must have a one-line summary (ending in .), followed by a blank line and then the rest of the description (https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings). All of the remaining errors come from either not following this convention (D205 and D400) or a missing docstring (D100, D101, D102, D103).

I also fixed a few typos while I was at it.


armenzg:
What a determination! Thanks Alice!

We should see what we need to do to not start loosing this increase of score. Eventually we should aim to make it a requirement to be compatible for future code commits.

vaibhavmagarwal:
Wohoo! This is great. Great Work @adusca

PR #133, opened 2015-03-13
Issue 117: gaia-try builders are always upstream.

adusca:


armenzg:
You rock! Thanks Alice!

PR #140, opened 2015-03-19
Refactored try in build_tests_per_platform_graph

adusca:
I replaced the try/except bit with a more specific condition, added a new test case and made platforms.py flake8-docstrigs compatible.


armenzg:
Thank you!

PR #153, opened 2015-04-05
[Outreachy] added a simple test to mozci.sources.allthethings.list_builders()

jasminu:
added a simple test to mozci.sources.allthethings.list_builders() using mock_allthethings.json as input data.

### DIFF
--- a/test/test_sources.py
+++ b/test/test_sources.py
@@ -1,0 +1,42 @@
+"""This file contains tests for mozci/mozci.py with data from mock_allthethings.json"""
+import json
+import os
+# import pytest
+import unittest
+
+from mock import patch
+
+import mozci.sources.allthethings
+
+
+def _get_mock_allthethings():
+    """Load a mock allthethings.json from disk."""
+    PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), "mock_allthethings.json")
+    with open(PATH, 'r') as f:
+        return json.load(f)
+
+MOCK_ALLTHETHINGS = _get_mock_allthethings()
+
+
+class TestSourcesAllTheThings(unittest.TestCase):
+
+    """Test allthethings with mock data."""
+
+    @patch('mozci.sources.allthethings.fetch_allthethings_data')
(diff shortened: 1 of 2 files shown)


adusca:
Hello Jaśmina! Thanks for your patch. Some suggestions:

I believe list_builders actually has a bug, namely, it's supposed to raise an AssertionError when the list of builders in empty, but it doesn't to that (because "[] is not None" evaluates as True). You could add a test with an empty list of builders (and fix the original bug) to check that case.

Because list_builders only uses the keys in the builders dictionary, instead of using mock_allthethings.json, I think it would be more clear if you did:
fetch_allthethingsdata.return_value = {'builders' : 
'Builder 1': {}, 
'Builder 2': {}}

Another advantage of doing it this way is that the test won't have to be updated whenever we edit mock_allthethings.json.

jasminu:
Hi Alice, thank you for your suggestions, I tried to address them with a new commit.

armenzg:
Hi Alice,
I've looked at the changes and they look fine. I think I would name the test test_allthethings.py but I would like to leave this to your judgement since you're more familiar with this part of the code than me.

Feel free to merge it if you believe it meets your expectations.

Thanks Jasmine!

test/test_sources.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class TestSourcesAllTheThings(unittest.TestCase):
+
+    """Test allthethings with mock data."""
+
+    @patch('mozci.sources.allthethings.fetch_allthethings_data')
+    def test_list_builders_with_mock_data(self, fetch_allthethings_data):
+        """is_dowstream should return True for test jobs and False for build jobs."""
+        fetch_allthethings_data.return_value = json.loads("""
```
adusca, line 17: You don't need to use json.loads here, you can just pass a dictionary. But this is just a matter of taste, you can keep it however you prefer.

adusca:
Thanks for updating your patch! Could you please fix the functions' docstrings? After you've done that and renamed the file I can merge your patch.

adusca:
Merged!

PR #161, opened 2015-04-09
Test fetch_allthethings_data

adusca:


test/test_allthethings.py
```diff
@@ -0,0 +1,138 @@
...
+import os
+import unittest
+
+from mock import patch, Mock
+
+from mozci.sources import allthethings
+
+
+TMP_FILENAME = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                            "tmp_allthethings.json")
```
armenzg, line 12: You can use the tempfile module.

NOTE: Take this (and following comments) into account if using the tempfile makes sense. Otherwise, please disregard!

test/test_allthethings.py
```diff
@@ -0,0 +1,138 @@
...
+            yield chunk
+
+    response.headers = {'content-length': str(len(data))}
+    response.iter_content = iter_content
+    return response
+
+
+class TestFetching(unittest.TestCase):
+
+    """Test fetch_allthethings_data() mocking requests."""
```
armenzg: Could you please add a more descriptive comment?

test/test_allthethings.py
```diff
@@ -0,0 +1,138 @@
...
+
+    allthethings.FILENAME = TMP_FILENAME
+    DATA = '{"data": 1}'
```
armenzg, line 40: Very smart :)

test/test_allthethings.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def setUp(self):
+        """Setting up values that will be used in every test."""
+        self.URL = "https://secure.pub.build.mozilla.org/builddata/reports/allthethings.json"
```
armenzg: You can use the constant ALLTHETHINGS from allthethings module.

test/test_allthethings.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def tearDown(self):
+        """Clean up after every test."""
+        if os.path.exists(TMP_FILENAME):
+            os.remove(TMP_FILENAME)
```
armenzg, line 51: If you use tempfile you won't need this.

test/test_allthethings.py
```diff
@@ -0,0 +1,138 @@
...
+        allthethings.DATA = None
+
+    @patch('requests.get', return_value=mock_get(DATA))
+    @patch('requests.head', return_value=Mock(headers={'content-length': str(len(DATA))}))
+    def test_calling_twice_with_caching(self, head, get):
+        """
+        We are going to call fetch_allthethings_data 2 times.
+
+        The first time it should use requests.get to download the file, and requests.head
+        to verify its integrity. The second time it will the file stored in-memory,
```
armenzg: Missing a word (I think). "It will read"

test/test_allthethings.py
```diff
@@ -0,0 +1,138 @@
...
+        assert get.call_count == 2
+        assert head.call_count == 2
+
+    @patch('requests.get', return_value=mock_get(DATA))
+    @patch('requests.head', return_value=Mock(headers={'content-length': str(len(DATA))}))
+    def test_calling_with_bad_cache(self, head, get):
+        """If the existing file is bad, we should download a new one."""
+        # Making sure the cache exists and it's bad
+        with open(TMP_FILENAME, 'w') as f:
+            f.write('bad file')
```
armenzg, line 94: If you use tempfile the file will be automatically deleted.

test/test_allthethings.py
```diff
@@ -0,0 +1,138 @@
...
+
+        self.assertEquals(allthethings.fetch_allthethings_data(), {'data': 1})
```
armenzg: Could you please define {'data': 1} variable as a constant at the top of the file?

test/test_allthethings.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def test_with_verify_set_to_false_and_no_cache(self):
+        """If verify is set to False and there is no file, it should raise an Error."""
+        with self.assertRaises(AssertionError):
+            allthethings.fetch_allthethings_data(verify=False)
+
+
+class TestSourcesAllTheThings(unittest.TestCase):
+
+    """Test mozci.sources with mock data."""
```
armenzg: Could you please add something more meaningful?
Change the class name if it makes sense to help know what we're testing down here.

adusca:
I addressed all non-tempfile-related comments, I'm working on tempfiles now.

adusca:
After experimenting with tempfile I believe it is not a good fit. It has good support for automatically deleting a file right after it is closed, but files that will be accessed more then once (like in this case) need to be manually deleted (and the final code is harder to understand).

I believe tempfile is going to be useful for testing buildjson.py

armenzg:
Thanks for trying it out. Thanks for addressing my comments.
If you're ready, please go ahead and merge.

PR #165, opened 2015-04-20
Added tests for internal functions in mozci.py.

adusca:


armenzg:
Wonderful! Thanks Alice!

Issue #168, opened 2015-04-30
Make it easier for developers to trigger jobs without giving buildername to mozci


vaibhavmagarwal:
We had a great discussion on IRC for this issue. I will summarize it here. Basically, developers need to figure out the buildername of a job for triggering the jobs via mozci. However, we can figure out the buildername for them, if they give us other useful information. Now, there are 2 ways to go about this:
1) They give us the Treeherder jobname and branch. For example, for "Linux opt mochitest-browser-chrome-1", we can have --jobname = "Linux opt Mochitest Mochitest M(1)" , --branch = "mozilla-inbound"

2) They give us the platform, testname and branch. Taking the same example as above, it will be:
--platform = "linux opt" --testname = "mochitest-browser-chrome-1" --branch = "mozilla-inbound"

@armenzg , @jmaher, @adusca which method do you prefer? I personally prefer the 2nd option, because it gives us scope to do add more functionality like if the developer specifies just: 
    --testname = "mochitest-browser-chrome-1" --branch = "mozilla-inbound"
then, we can trigger that test on all the platforms (i.e. desktop, android, b2g) which might not be possible in the first method.

If we go with the second method, we would need to create some sort of dynamic mapping to filter the buildernames in steps.

adusca:
For option 2 we could just add something like that to platforms.py:

def find_buildername(test, repo, platform):

right_tests = _filter_builders_matching(fetch_allthethings_data().keys(), test)

repo_and_test =_filter_builders_matching(right_tests, repo)

for buildername in repo_and_test:

if get_associated_platform_name(buildername) == platform:

return buildername

armenzg:
2nd method approach in my opinion should be better. We can start with a POC and re-think if need be.

jmaher:
I would say, make a way to default values for them, specifically repo- you either land on inbound or fxteam.  'linux opt' -> I would prefer linux, and assume opt.

for the case of try server, we should either make it easy to do or default there.  I suspect most developers will want this on try, so maybe the default repo is try.

friendlyretrigger.py --rev <asfd> --testname mochitest-other

I think requiring:
--testname mochitest-devtools-2
or
--testname mochitest-e10s-devtools-2

is bad.  we should make it easy:
--test mochitest --test devtools  <- defaults to all mochitest devtools
--test dt <- translates to mochitest-devtools (but not the chunks)
--test bc1 <- mochitest-browser-chrome-1, mochitest-browser-chrome-e10s-1

[...]

PR #169, opened 2015-05-01
Added function find_buildernames to help with issue 168

adusca:


vaibhavmagarwal:
Awesome @adusca . I think that "function find_buildernames" is a bit too specific and should not be in platforms.py, but we can move it into a separate script afterwards. And nice removal of extra trailing spaces.

mozci/platforms.py
```diff
@@ -257,3 +257,30 @@ def build_talos_buildernames_for_repo(repo_name, pgo_only=False):
 
     retVal.sort()
     return retVal
+
+
+def find_buildernames(repo, test=None, platform=None):
```
armenzg, line 262: I'm not suggesting to change the name but "filtered_builders" could also be another name for it if "find" does not work out.

mozci/platforms.py
```diff
@@ -257,3 +257,30 @@ def build_talos_buildernames_for_repo(repo_name, pgo_only=False):
...
+    """
+    Return a list of buildernames matching the criteria.
+
+    1) if the developer provides test, repo and platform, return only the specific buildername
+    2) if the developer provides test and platform only, then return the test on all platforms
+    3) if the developer provides platform and repo, then return all the tests on that platform
+    """
+    if test is None and platform is None:
```
armenzg, line 270: You can have an assertion in here
assert true, "Message to be raised"

mozci/platforms.py
```diff
@@ -257,3 +257,30 @@ def build_talos_buildernames_for_repo(repo_name, pgo_only=False):
...
+    matching = []
+    buildernames = _filter_builders_matching(fetch_allthethings_data()['builders'].keys(), repo)
+
+    if test is not None:
+        buildernames = _filter_builders_matching(buildernames, test)
+
+    if platform is not None:
+        for buildername in buildernames:
+            if get_associated_platform_name(buildername) == platform:
+                if is_downstream(buildername):
```
armenzg, line 282: Would this work _only_ with downstream jobs? (aka test jobs)

armenzg:
You're such a pro at writing tests and understanding business needs. Well done!

Please merge once the comments are addressed.

adusca:
Thanks!

If I removed "if is_downstream(buildername):" we would get every test job and a build job when the test is not specified. I asked @vaibhavmagarwal and he believes having only test jobs fits #168 better, because the build job would be automatically triggered if no suitable one is found. What do you think?

armenzg:
Sure!
As long as we're aware of what we're doing.
If we need to change in the future we can.

PR #173, opened 2015-05-04
Fixed assertion, added spaces to find_buildernames

adusca:
I misunderstood a previous comment about adding an assertion and ended up merging something wrong. This fixes it.

I also added spaces around repo to avoid false positives (so we don't get e.g. gaia-try builds when looking for try builds)

### DIFF
--- a/mozci/platforms.py
+++ b/mozci/platforms.py
@@ -162,7 +162,7 @@ def _get_test(buildername):
 
 def _filter_builders_matching(builders, keyword):
     """Find all the builders in a list that contain a keyword."""
-    return filter(lambda x: keyword in x, builders)
+    return map(str, filter(lambda x: keyword in x, builders))
 
 
 def build_tests_per_platform_graph(builders):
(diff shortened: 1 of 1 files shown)


adusca:
Turns out that only looking at the platform name does not tell us if a build is debug or not (test jobs do not have -debug in their 'platform' field, only build jobs do):

> > > from mozci.platforms import find_buildernames
> > > find_buildernames('try', 'mochitest-1', 'macosx64')
> > > ['Rev5 MacOSX Yosemite 10.10 try debug test mochitest-1', 'Rev5 MacOSX Mountain Lion 10.8 try opt test mochitest-1', 'Rev4 MacOSX Snow Leopard 10.6 try opt test mochitest-1', 'Rev4 MacOSX Snow Leopard 10.6 try debug test mochitest-1', 'Rev5 MacOSX Mountain Lion 10.8 try debug test mochitest-1', 'Rev5 MacOSX Yosemite 10.10 try opt test mochitest-1']
> > > find_buildernames('try', platform='macosx64-debug', test='mochitest-1')
> > > []

I fixed this by checking if ' debug ' is in the buildername. I bel

PR #179, opened 2015-05-07
Add a parameter for extra buildbot properties to trigger functions.

chmanchester:


scripts/trigger.py
```diff
@@ -109,6 +109,7 @@ def validate_options(options):
     if error_message:
         raise Exception(error_message)
 
+
 def sanitize_buildername(buildername):
```
vaibhavmagarwal, line 113: This is no longer required. @adusca has merged this in https://github.com/armenzg/mozilla_ci_tools/pull/181

mozci/sources/buildapi.py
```diff
@@ -82,13 +83,19 @@ def _jobs_api_url(job_id):
     return r'''%s/jobs/%s''' % (HOST_ROOT, job_id)
 
 
-def _payload(repo_name, revision, files=[]):
-    payload = {}
-    # These properties are needed for Treeherder to display running jobs
-    payload['properties'] = json.dumps({
+def _payload(repo_name, revision, files=[], extra_properites=None):
+
```
vaibhavmagarwal, line 87: @chmanchester, could you please remove the extra line here. Also, please change the comments from having # to a string as in other functions, just to be consistent. Thanks :)

armenzg:
Just what @vaibhavmagarwal mentioned.
I will merge it as soon as you push the style change.

Don't worry about the coveralls results. We're mainly using it to track where we stand.

mozci/sources/buildapi.py
chmanchester, line 87: Docstrings are intended to be read by documentation generating tools. This is an internal function and the comment doesn't pertain to the entire function, so I think an inline comment if more appropriate.

vaibhavmagarwal, line 87: Oh Okay. I thought it was for the whole function. So lets keep it as it is. Thanks for the clarification.

PR #180, opened 2015-05-07
Add function _get_job_type. Add parameter job_type to find_buildernames

adusca:
Based on discussion in #168

### DIFF
--- a/mozci/platforms.py
+++ b/mozci/platforms.py
@@ -160,6 +160,23 @@ def _get_test(buildername):
     return buildername.split(" ")[-1]
 
 
+def _get_job_type(test_job):
+    """
+    Classify a job as 'opt', 'debug' or 'pgo' based on its name.
+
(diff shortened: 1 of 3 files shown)


mozci/platforms.py
```diff
@@ -160,6 +160,23 @@ def _get_test(buildername):
     return buildername.split(" ")[-1]
 
 
+def _get_job_type(test_job):
+    """
+    Classify a job as 'opt', 'debug' or 'pgo' based on its name.
+
+    Currently only working for test jobs.
```
armenzg: We should add a note that this only works for Buildbot's desktop and mobile jobs naming.
This function does not apply to B2g and TaskCluster.

armenzg:
Alice this looks very nice. Thanks for fixing it!

armenzg:
Please merge once you add that comment. Thanks!

PR #198, opened 2015-05-14
Add a way to filter out android/desktop and/or specific talos jobs

gijsk:
I needed to do this for my perf analysis and I felt bad retriggering all these suites/platforms I didn't need.

### DIFF
--- a/mozci/scripts/alltalos.py
+++ b/mozci/scripts/alltalos.py
@@ -49,6 +49,17 @@ def parse_args(argv=None):
                         dest="pgo",
                         help="trigger pgo tests (not non-pgo).")
 
+    parser.add_argument("--filter",
+                        action="store",
+                        dest="filterstring",
+                        help="comma-separated substring filter indicating which jobs to retrigger")
+
+    parser.add_argument("--platform",
+                        action="store",
+                        default="all",
+                        choices=["desktop", "android", "all"],
+                        help="indicate which platform to run jobs for (default: all)")
+
     options = parser.parse_args(argv)
     return options
 
@@ -71,7 +82,35 @@ def main():
 
     buildernames = build_talos_buildernames_for_repo(options.repo_name, pgo)
 
+    jobfilters = None
+    if options.filterstring:
+        jobfilters = options.filterstring.split(',')
+
     for buildername in buildernames:
+        try:
+            buildername.index("ndroid")
+            # This is an android build; if we only want desktop, ignore:
+            if options.platform == "desktop":
+                continue
+        except:
+            # This is a desktop build; if we only want android, ignore:
+            if options.platform == "android":
+                continue
+
+        if jobfilters:
+            shouldrun = False
+            for jobfilter in jobfilters:
+                try:
+                    buildername.index(jobfilter)
+                    shouldrun = True
+                    break
+                except:
+                    pass
+
+            if not shouldrun:
+                LOG.info("Not triggering " + buildername + " due to filter.")
+                continue
+
         trigger_job(revision=options.revision,
                     buildername=buildername,
                     times=options.times,


mozci/scripts/alltalos.py
```diff
@@ -71,7 +82,35 @@ def main():
...
+        except:
+            # This is a desktop build; if we only want android, ignore:
+            if options.platform == "android":
+                continue
+
+        if jobfilters:
+            shouldrun = False
+            for jobfilter in jobfilters:
+                try:
+                    buildername.index(jobfilter)
```
jmaher, line 104: my only concern here is a typo, I guess the info message below would accommodate that.

I know there is a lot of work which has been done in the trigger scripts which falls into this filtering category.  I think this is a good change to start with for the talos bits, eventually we can have a more comprehensive selection/filtering toolchain that works on all builds/tests/talos jobs.

armenzg:
@adusca recently added some Treeherder filtering functionality to one of the other script.
We should streamline it and add it in here.

Happy if we take care of fixing the issue?

jmaher:
the work adusca has done is in triggerbyfilters.py, does that work for talos?  If so we could maybe deprecate alltalos.py an instead:
python triggerbyfilters.py --rev xyz --branch try --filter talos,svgr --times 6

if we wanted to run two different talos suites, we would have to:
python triggerbyfilters.py --rev xyz --branch try --filter talos,svgr --times 6
python triggerbyfilters.py --rev xyz --branch try --filter talos,other --times 6

The last thing would be to filter out desktop|android|all jobs.

adusca:
@jmaher tryggerbyfilters.py currently doesn't deal with mozilla-aurora and mozilla-beta special pgo talos cases

jmaher:
For what Gijs is doing, he is looking at Try server (think opt only).  This
might make it more plausible!

adusca:
For try, this should work:

python triggerbyfilters.py try REVISION --includes 'talos svgr' --times 6

jmaher:
one question is there are a couple other tests:
talos other
mochitest-other

if we have --includes 'talos other', will it trigger all jobs matching
talos + other?

if we have --includes 'talos' will it trigger ALL talos jobs?

I assume we ignore pgo builders with this, right?

btw, this is great stuff!

adusca:
--includes 'talos other' takes every job that has both  'talos' and 'other' in its name, so by default it will take pgo to.
If you want to exclude pgo builder you can do: --includes 'talos other' --exclude 'pgo'

gijsk:
It is not clear to me how this is meant to work because I don't know enough about these scripts.

I was originally warned that for the talos script, it will trigger builds for architectures where the requisite builds are not available. Is this the case for the other script as well? Because that would seem suboptimal (ie I would not want to accidentally be triggering extra PGO builds for all the platforms on the push, or have to specify `--exclude 'pgo'` every time I called the script).

armenzg:
This is due to the nature of try being a repo which triggers jobs depending on a try syntax.
Using filters can trigger jobs which the try syntax would have not.

What was your try push?
What did you want to add?

Perhaps the better use of our time is to fix:
http://mozilla-ci-tools.readthedocs.org/en/master/use_cases.html#case-scenario-8-developer-needs-to-add-missing-platforms-jobs-for-a-try-push

and/or build a web tool to allow to visually check/uncheck platforms you want triggered (a web view of that use case).

jmaher:
this specific use case is pushing a talos fix and wanting to retrigger specific jobs X times.  When you have 3+ try pushes, a script becomes a life saver (at least a good time saver).

The danger with some scripts is we end up triggering android jobs or other platforms which might not be affected by the changes in question.  I believe Gijs was trying to do the right thing by selectively retriggering jobs so as not to waste resources.

armenzg:
We should add --exclude 'pgo' by default, specially for try.

The filtering functionality is being upstreamed and will be available for this script on:
https://github.com/armenzg/mozilla_ci_tools/pull/201

As mentioned, I agree that we should look into deprecating alltalos.py if it makes sense.

I've filed https://github.com/armenzg/mozilla_ci_tools/issues/206 in order to keep the discussion open but we able to close the PR in itself.

jmaher:
on try there is no PGO, so there would be no worries there.  Still exclude
pgo should be on by default. and we should make it --pgo or --include 'pgo'

PR #199, opened 2015-05-14
Better wording when triggering builds instead of test jobs

armenzg:
This should help people understand what is going.


adusca:
r+

Issue #188, opened 2015-05-12
Do not fix libraries versions in the setup.py


parkouss:
Hm, I have a strange bug. I have one application that is using mozci and bugsi: https://github.com/parkouss/mozbattue/blob/master/setup.py

Pip can not install these properly (I tried pip 6.0.7 and latest 6.1.1): https://travis-ci.org/parkouss/mozbattue/builds/62225972

It seems that if I run "pip install mozci" after the issue, all is running fine. I looked into mozci's setup.py script, and there is bugsy listed as a dependency (with a fixed version).

This seems like a pip bug - still I'm not 100% sure as I did not checked every requirement of every dependencies recursively.

However I think that it may be good to not fix versions this way in mozci - as it is intended to be used as a library, this could easily conflict with other libraries dependencies or one of the program dependencies itself. Maybe something like "bugsy>=0.4" and "requests>=2.5.1" would be a good solution? This is a proposal, and I can understand if you find this too risky. :)

I think that should fix my problem - note that I could also do not make bugsy a dependency, and hope that future mozci versions will still use bugsy. This is an open-to-discussion issue!

armenzg:
Do you want to give it a shot and let me know how it goes?

I think we could loosen up the requirements if it proves to help you.

parkouss:
Hm, this really seems to be a pip bug somewhere, or at least something I don't understand now. I am not able to reproduce it with
pip install mozci bugsy && pip install -e .
but only when I run pip install -e .
(each command in a fresh virtualenv, done in my cloned dir). So this is really strange, and I'm not sure that not fixing the version would help now. I think that for my need here, I will only make a dependency on mozci (this works for each command) - and I will inherit the bugsy dependency anyway.

So you can close this issue - unless you'd want to take the opportunity to not fix the versions.

vaibhavmagarwal:
I think older versions of requests and pip have conflicts.
@parkouss Same error as you got: https://github.com/docker/compose/issues/918
Also check: https://github.com/pypa/pip/issues/775

vaibhavmagarwal:
I think this is really not a mozci problem, at most we can upgrade requests version in setup.py

parkouss:
Thanks @vaibhavmagarwal for looking into this. :) So yes, sounds like a pip bug;

I suspect that if "requests >= 2.5.1" is specified in mozci, it won't fail anymore for me - but it won't work if you specify something like "requests == 2.5.3".

[...]

PR #204, opened 2015-05-14
Issue 188 - Loosen up the python requirements for mozci

armenzg:
To fix @parkouss' #188

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,11 +13,11 @@ setup(
         ],
     },
     install_requires=[
-        'beautifulsoup4==4.3.2',
-        'bugsy==0.4.0',
-        'progressbar==2.3',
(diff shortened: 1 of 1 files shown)


armenzg:
r? @vaibhavmagarwal 

Should we remove bugsy as a core dependency of mozci? It seems that we should have a try/block in generatecli and ask the dev to install bugsy to run that script. Thoughts?

vaibhavmagarwal:
r+ as it is.
I vote for having bugsy as a core dependency and not as a try/block in generatecli, because it is better to install every library with 'python setup.py develop' rather than to add another step for the developer, and make them install it manually.

Also, loosing up the dependencies like this, solves the problem @parkouss was facing. Merging this pr.

PR #205, opened 2015-05-15
Fix gzip issues on Windows

armenzg:
The gzip.py python module throws MemoryError when decompressing very big files.
Instead we're switching to use the gzip binary to decompress the files on Windows.
This makes it necessary that gzip is installed on your system.
If you install mozilla-build you should be fine.

### DIFF
--- a/mozci/utils/transfer.py
+++ b/mozci/utils/transfer.py
@@ -2,11 +2,8 @@ import gzip
 import json
 import logging
 import os
-import shutil
 import time
 import StringIO
-
-from tempfile import NamedTemporaryFile
 
 import requests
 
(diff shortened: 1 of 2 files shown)


mozci/utils/transfer.py
```diff
@@ -57,28 +59,56 @@ def _fetch_and_load_file(req, filename):
...
+            cmd = ["gzip", "-d", filename + ".gz"]
+            LOG.debug("-> %s" % ' '.join(cmd))
+            try:
+                retcode = subprocess.call(cmd)
+            except WindowsError, e:
+                if str(e) == "[Error 2] The system cannot find the file specified": 
+                    raise MozciException(
+                        "You don't have gzip installed on your system. "
+                        "Please install it. You can find it inside of mozilla-build."
+                    )
```
vaibhavmagarwal, line 73: Any specific reason to raise MozciException? Can we just raise Exception here?

vaibhavmagarwal:
Tox is failing here. Looks like your editor is mixing up tabs and spaces

vaibhavmagarwal:
I haven't tested it on my windows machine as it is not with me right now, but the general logic looks right.
r+ with the nits fixed.

armenzg:
Remind me not to write and commit code on Windows :)
I will fix this after lunch.

mozci/utils/transfer.py
armenzg, line 73: Yep.

PR #215, opened 2015-05-19
Fixed query_repo_name_from_buildername wrongly matching branchs

adusca:
I noticed sooner that: 

> > > query_repo_name_from_buildername('Linux try build')
> > > u'ux'

This patch fixes that:

> > > query_repo_name_from_buildername('Linux try build')
> > > u'try'


armenzg:
Oh wow! Good catch there!

PR #217, opened 2015-05-19
Explained in the logs what happens with build jobs/test jobs.

adusca:
Output: https://pastebin.mozilla.org/8833980


armenzg:
Much better worded! Thanks Alice.

PR #220, opened 2015-05-20
Issue 219: Look for files in unsucessful jobs, don't retrigger failed builds

adusca:
Example with failed build job without files:
https://pastebin.mozilla.org/8834105

Failed job with files:
https://pastebin.mozilla.org/8834109
(it's a testfailed job: https://treeherder.mozilla.org/#/jobs?repo=try&fromchange=e212664d709b&tochange=e212664d709b )

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -112,11 +112,22 @@ def _determine_trigger_objective(revision, buildername):
 
     # We need to determine if we need to trigger a build job
     # or the test job
-    successful_job = None
+    working_job = None
     running_job = None
+    failed_job = None
 
     LOG.debug("List of matching jobs:")
     for job in build_jobs:
+        # Successful, running and failed jobs may have the files we need
+        files = _find_files(job)
+        if files != [] and _all_urls_reachable(files):
+            working_job = job
+            break
+        else:
+            LOG.debug("We can't determine the files for this build or "
+                      "can't reach them.")
+            files = None
+
         try:
             status = buildapi.query_job_status(job)
         except buildjson.BuildjsonException:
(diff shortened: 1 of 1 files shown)


mozci/mozci.py
```diff
@@ -112,11 +112,22 @@ def _determine_trigger_objective(revision, buildername):
...
     # or the test job
-    successful_job = None
+    working_job = None
     running_job = None
+    failed_job = None
 
     LOG.debug("List of matching jobs:")
     for job in build_jobs:
+        # Sucessful, running and failed jobs may have the files we need
+        files = _find_files(job)
```
armenzg, line 122: Running jobs might have uploaded files, however, I don't know if the properties are updated on the buildbot side and we might not be able to find the files until completed.
This is OK and as I said "I don't know" if the files would be available as properties before completion. Who knows, maybe it works.

mozci/mozci.py
```diff
@@ -127,34 +138,26 @@ def _determine_trigger_objective(revision, buildername):
...
+            LOG.info("We found a job that finished but its status "
+                     "is not successful. status: %d" % status)
+            failed_job = job
 
-    if successful_job:
-        # A build job has completed successfully and the files can be reached
-        LOG.info("There is a _build_ job that has completed successfully.")
-        LOG.debug(str(successful_job))
+    if working_job:
+        # The files can be reached
```
armenzg: Could you please add a note of what a working job is?

armenzg:
This is great. I think only jmaher and few people will understand what a big and better behaviour this is,ho wever, I'm glad you figured it out as this will make the flow more smooth.

I'm not going to ask to add a --force-build parameter until we actually have a customer that would want a different behaviour.

adusca:
I added a comment to explain what a working job is.

PR #221, opened 2015-05-21
Caching jobs

adusca:
"python mozci/scripts/triggerbyfilters.py try 443657964b30 -i 'mochitest-1' -e android,b2g,pgo,10.8,coverage,asan --dry-run --times 7"
takes 26 seconds in this branch (versus 131 seconds in master)

I believe this change is worth merging.

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -21,7 +21,7 @@ from mozci.utils.transfer import path_to_file
 
 LOG = logging.getLogger('mozci')
 SCHEDULING_MANAGER = {}
-
+JOBS_CACHE = {}
 
 def _matching_jobs(buildername, all_jobs):
     """Return all jobs that matched the criteria."""
@@ -215,7 +215,10 @@ def _find_files(job_schedule_info):
 #
 def query_jobs(repo_name, revision):
     """Return list of jobs scheduling information for a revision."""
-    return buildapi.query_jobs_schedule(repo_name, revision)
+    global JOBS_CACHE
+    if (repo_name, revision) not in in JOBS_CACHE:
+        JOBS_CACHE[(repo_name, revision)] = buildapi.query_jobs_schedule(repo_name, revision)
+    return JOBS_CACHE[(repo_name, revision)]
 
 
 def query_jobs_buildername(buildername, revision):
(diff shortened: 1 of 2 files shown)


mozci/mozci.py
```diff
@@ -215,7 +216,14 @@ def _find_files(job_schedule_info):
...
 def query_jobs(repo_name, revision):
     """Return list of jobs scheduling information for a revision."""
-    return buildapi.query_jobs_schedule(repo_name, revision)
+    global JOBS_CACHE
+    if (repo_name, revision) not in JOBS_CACHE:
+        # We currently have 2 use cases with multiple jobs: triggering
+        # several jobs in one revision or triggering the same job in
+        # different revisions. Caching query_jobs is only useful for
+        # the first case, and for that we only need to cache one job
+        JOBS_CACHE = {(repo_name, revision): buildapi.query_jobs_schedule(repo_name, revision)}
```
vaibhavmagarwal: I think here you should add the (repo_name, revision) to the existing dictionary rather than to make a separate dictionary each time for every revision. This is because if we trigger across multiple revisions (with say --delta 2), the JOBS_CACHE can still be used optimally. So, in my opinion:

```
JOBS_CACHE[(repo_name, revision)] = buildapi.query_jobs_schedule(repo_name, revision)
```

would give better results for a range of revisions. Though I would like a benchmark to make sure the change I suggest makes sense.

armenzg: This is a good idea, however, we should put this logic inside of buildapi.py rather than mozci.py
Once I work on TC support, I _might_ move some functionality on mozci.py into other modules (I might need to move the SCHEDULING_MANAGER).

Let us know how @vaibhavmagarwal 's suggestion works out.

armenzg:
This is great Alice.
Please poke me once more after you address the comments.

adusca:
I moved all the caching to buildapi.py. I also removed the restriction in JOBS_CACHE's size, so it can now cache more than one value.

armenzg:
I missed this! I will review this after I review bug 1168148.
On GH, please type "r?" in a new line on a comment or make reference to my nickname.
Hopefull that will catch my eye and won't miss it. Maybe! I don't know

With bugzilla is a bit better since I can see a nagging red circle on my bugzilla showing that I have something pending. With GH I only see a blue circle above the inbox but it can refer to any activity so I never check it. I only use emails sent from GH and I don't know how I missed this. Sorry!

PR #226, opened 2015-05-29
Issue 225: Properly cache downloaded files

glandium:


adusca:
Good catch! Thanks

armenzg:
We should do a dot release to contain this.

armenzg:
Thanks @glandium !

Issue #121, opened 2015-03-12
Alltalos should look at the try syntax OR at the list of existing build jobs to determine which downstream jobs we need to trigger (Only for try)


glandium:
Related to issue #120, what I initially wanted to do is trigger 2 runs of every talos job *on the build that was already there, and not any other*.

armenzg:
If this gets fixed with updating the checkout please let ys know.

glandium:
This is still an issue:
05/29/2015 11:28:08 INFO:	 We want to trigger 'Android 4.0 armv7 API 11+ try talos remote-tp4m_nochrome' on revision '0c31667c79fa' a total of 5 time(s).
05/29/2015 11:28:49 INFO:	 We will trigger 1) 'Android armv7 API 11+ try build' instead of 2) 'Android 4.0 armv7 API 11+ try talos remote-tp4m_nochrome'
05/29/2015 11:28:49 INFO:	 We need to trigger the build job once (1) in order to be able to run the test job (2).
05/29/2015 11:28:49 INFO:	 You'll need to run the script again after (1) is done to trigger (2).
05/29/2015 11:28:49 INFO:	 In order to trigger Android 4.0 armv7 API 11+ try talos remote-tp4m_nochrome 5 times, please run the script again after Android armv7 API 11+ try build ends.
05/29/2015 11:28:49 INFO:	 Dry-run: We were going to request 'Android armv7 API 11+ try build' 1 times.
05/29/2015 11:28:49 INFO:	 Dry-run: We were going to request a job for 'Android armv7 API 11+ try build'

armenzg:
I'm looking at https://treeherder.mozilla.org/#/jobs?repo=try&revision=0c31667c79fa
I only see Linux x64 jobs there, however, in your output you're talking about Android jobs.
I'm going to change the name of this issue to help me not get confused in the future.

For non-try repos, we should not follow this behaviour since, sometimes, there are builds which are skipped.

Poking @jmaher to be in the loop.

armenzg:
I'm leaning towards the list of build jobs on a try push over the try syntax since we might have try pushes where developers have added more platforms post-push (i.e. for instance through mozci-trigger).

glandium:
> I'm looking at https://treeherder.mozilla.org/#/jobs?repo=try&revision=0c31667c79fa
> I only see Linux x64 jobs there, however, in your output you're talking about Android jobs.

[...]

PR #228, opened 2015-05-29
Issue #121: added --trigger-build-if-missing flag to alltalos.py

adusca:
Example output of running "python mozci/scripts/alltalos.py --repo-name try --dry-run --existing-only --rev  b71c4e5e9542 --times 2" in a revision with android jobs only: https://pastebin.mozilla.org/8835070

We will need to clean up the logs a bit, but I believe this can be left for a later patch.

I had to move the SCHEDULING_MANAGER update to trigger() because _unique_build_request assumed checking for uniqueness implied triggering a build, which is not the case here.

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -48,9 +48,9 @@ def _unique_build_request(buildername, revision):
         return True
     else:
         if revision in sch_mgr and buildername in sch_mgr[revision]:
-            LOG.info("We have already scheduled the build '%s' for "
-                     "revision %s during this session. We don't allow "
-                     "multiple requests." % (buildername, revision))
+            LOG.debug("We have already scheduled the build '%s' for "
+                      "revision %s during this session. We don't allow "
+                      "multiple requests." % (buildername, revision))
             return False
         return True
 
@@ -77,7 +77,7 @@ def _status_summary(jobs):
     return (successful, pending, running, coalesced)
 
 
-def _determine_trigger_objective(revision, buildername, existing_only=False):
+def _determine_trigger_objective(revision, buildername, trigger_build_if_missing=True):
     """
     Determine if we need to trigger any jobs and which job.
 
@@ -159,7 +159,7 @@ def _determine_trigger_objective(revision, buildername, existing_only=False):
     else:
         # We were trying to build a test job, however, we determined
         # that we need an upstream builder instead
-        if existing_only or not _unique_build_request(build_buildername, revision):
+        if not trigger_build_if_missing or not _unique_build_request(build_buildername, revision):
             # This is a safeguard to prevent triggering a build
             # job multiple times if it is not intentional
             builder_to_trigger = None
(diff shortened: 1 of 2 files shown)


glandium:
If you pass existing_only down to _determine_trigger_objective, you can replace `if not _unique_build_request(build_buildername, revision)` with `if existing_only or not _unique_build_request(build_buildername, revision)`, and that avoids all the extra logging.

glandium:
(btw, I would argue that this --existing-only behavior should be the default)

armenzg:
@adusca for try, we should look at the data from buildapi to determine which jobs have been scheduled and only consider those builds to add more talos jobs to.

For instance, on that revision:
https://treeherder.mozilla.org/#/jobs?repo=try&revision=b71c4e5e9542
we should only find which of the upstream builders have any talos downstream jobs:
 Android 2.3 API9 opt Build (B) 
 Android 4.0 API11+ opt Build (B) 
 Android 4.2 x86 opt Build (B) 

We should not go consider Mac, Windows or Linux for this specific revision.

Makes sense to you?

adusca:
@glandium's suggestion cleaned up the logs a bit, but we still log every job we considered but won't trigger: https://pastebin.mozilla.org/8835127

This makes --existing-only mandatory for try in alltalos.

r?

mozci/mozci.py
```diff
@@ -52,12 +52,7 @@ def _unique_build_request(buildername, revision):
                      "revision %s during this session. We don't allow "
```
armenzg: Should this LOG.info() message be a LOG.debug()?
I don't think developers' care.

mozci/mozci.py
```diff
@@ -350,6 +345,7 @@ def trigger_job(revision, buildername, times=1, files=None, dry_run=False,
         builder_to_trigger, files = _determine_trigger_objective(
             revision,
             buildername,
+            existing_only
```
armenzg: Could you change `existing_only` in where appropiate for `trigger_build_if_missing`? It's more clean, I think.

mozci/scripts/alltalos.py
```diff
@@ -49,6 +49,11 @@ def parse_args(argv=None):
                         dest="pgo",
                         help="trigger pgo tests (not non-pgo).")
 
+    parser.add_argument("--existing-only",
+                        action="store_true",
+                        dest="existing_only",
+                        help="only retrigger jobs for which there is an existing build")
```
armenzg: Could you change this according to my previous request to rename `existing_only`?

armenzg:
Thanks Alice. This is looking good. The main change is the renaming of that variable and related flag.

adusca:
Renaming --existing-only to --trigger-build-if-missing changed the behaviour a little bit:
trigger_build_if_missing is False by default on try and True by default on every other repo. Running with --trigger-build-if-missing will set trigger_build_if_missing as True, so it makes no difference in other repos and changes the behaviour on try.

(Previously, --existing-only made no difference on try and changed the behaviour in other repos)

Issue #227, opened 2015-05-29
_fetch_and_load_file is only caching compressed downloads, in uncompressed form


glandium:
In b584fcc087340cc1b2d1fa640841354973c4ac42, I removed one level of indentation that 5e9f2cecf5b7cb1102505f32b4876d6b192a0be2 added. Looking further, that was already one level of indentation too much before that: the cache is only stored if the requested file has x-gzip Content-Encoding. It also only does that only for non-windows, so Windows does not cache at all.
The cache is also stored uncompressed, which for builds-4hrs.gz is 30M instead of 3M.
And in fact, _fetch_and_load_file will throw an UnboundLocalError exception when the downloaded file doesn't have a x-gzip Content-Encoding because json_content is only set if it does.

glandium:
Let me clean this up after lunch.

adusca:
Fixed in #229.

PR #229, opened 2015-05-29
Issue #227: Cache files in the same form they were on the wire

glandium:
Caching gzipped files in uncompressed form is a waste of disk space, so
store them in gzipped form instead. At the same properly cache on Windows
and when the downloaded file was not gzipped, which previously didn't lead
to data being cached.

Note that this means the file is gunzipped every time it's used, when it previously was read uncompressed every time, which is not necessarily better. The right thing to do would be to cache the json data in memory, but I'll leave that up to you.

### DIFF
--- a/mozci/utils/transfer.py
+++ b/mozci/utils/transfer.py
@@ -1,3 +1,4 @@
+import errno
 import gzip
 import json
 import logging
@@ -6,7 +7,6 @@ import platform
 import shutil
 import subprocess
 import time
-import StringIO
 
 import requests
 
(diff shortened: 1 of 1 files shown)


mozci/utils/transfer.py
```diff
@@ -121,27 +101,46 @@ def load_file(filename, url):
...
+            for chunk in req.iter_content(10 * 1024):
+                if chunk:  # filter out keep-alive new chunks
+                    fd.write(chunk)
+                    bytes += len(chunk)
+                    pbar.update(bytes)
+        pbar.finish()
+
+    elif req.status_code == 304:
+        # The file on disk is recent
+        LOG.debug("%s is on disk and it is current." % last_mod_date)
```
armenzg, line 141: This message talks about file yet we subsitute with a date.

armenzg:
Thanks glandium. My apologies about the delay. GH does not have reminders like BZ (afaik).

I've thought about it and for most cases we should not have a big performance hit. We only load a small number of those files at a time.

I would like to test this few times before I merge it in.

armenzg:
We are fetching builds-4hr multiple times when we should not (The file only changes once every minute and we fetched it several times in few seconds).

We store the file with .js extension (even thought it is a gzip file):
-rw-r--r--  1 armenzg  staff   4283654  3 Jun 14:14 builds-4hr.js

The pastebin shows 2 fetches of builds-4hr.js.gz:
https://pastebin.mozilla.org/8835582

Here's relevant output:
06/03/2015 02:14:10 DEBUG:   We have not been able to find /Users/armenzg/.mozilla/mozci/builds-4hr.js on disk.
...
06/03/2015 02:14:17 DEBUG:   "GET /builddata/buildjson/builds-4hr.js.gz HTTP/1.1" 200 4283654
06/03/2015 02:14:17 DEBUG:   The local file was last modified at Wed, 03 Jun 2015 18:14:14 GMT. We need to fetch it again.

mozci/utils/transfer.py
glandium, line 141: > This message talks about file yet we subsitute with a date.

I didn't change that line ;) It comes from the current file.

glandium:
> We store the file with .js extension (even thought it is a gzip file):

That simplified the code to do so.

> The pastebin shows 2 fetches of builds-4hr.js.gz:

The part that takes the file timestamp and requests that hasn't changed, so this wouldn't be a new problem. Can you add a log of the last-modified from the response headers?

glandium:
That said, the time difference between the log times and the reported last modified time is weird. You're not in GMT-16 are you? ;) So there's something funky with the timezone handling, but that must have been there before.

armenzg:
I will investigate this and merge together with any new findings.

armenzg:
I've fixed the issue that I noticed:
https://github.com/armenzg/mozilla_ci_tools/commit/1e591bfc62e3f1afdc2a10d198208363bf666766

I will merge this in a bit.

armenzg:
I've fixed the merge conflicts and have taken this change into "store_as_gzip" branch so we test it for few days before making it part of the main release.

Thanks @glandium !

armenzg:
I've made some changes, tested it and pushed it.
Let's way few days before we do a release.

PR #234, opened 2015-06-01
Issue 232: Don't try to find files in running jobs

adusca:
I discovered that sometimes running jobs have status 'UNKNOWN'. I'm dealing with that by assuming that by treating 'UNKNOWN' as 'RUNNING', I don't know if there is a better way.


armenzg:
Thank you Alice!

PR #241, opened 2015-06-03
Issue 240: Allow multiple buildernames in trigger.py

adusca:
Output of running python mozci/scripts/trigger.py -b "Ubuntu VM 12.04 try opt test mochitest-1","Ubuntu VM 12.04 try opt test mochitest-2" -r 74088ab66f44 --dry-run: https://pastebin.mozilla.org/8835618

Output of running python mozci/scripts/trigger.py -b "Ubuntu VM 12.04 try opt test mochitest-1" -r 74088ab66f44 --dry-run: https://pastebin.mozilla.org/8835619

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -386,10 +386,7 @@ def trigger_job(revision, buildername, times=1, files=None, dry_run=False,
 
 
 def trigger_range(buildername, revisions, times=1, dry_run=False, files=None):
-    """
-    Schedule the job named "buildername" ("times" times) from "start_revision" to
-    "end_revision".
-    """
+    """Schedule the job named "buildername" ("times" times) in every revision on 'revisions'."""
     repo_name = query_repo_name_from_buildername(buildername)
     LOG.info("We want to have %s job(s) of %s on revisions %s" %
              (times, buildername, str(revisions)))
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -468,7 +468,7 @@ def trigger(builder, revision, files=[], dry_run=False, extra_properties=None):
                                           extra_properties)
 
 
-def backfill_revlist(buildername, revisions, times=1, dry_run=False):
+def backfill_revlist(buildername, revisions):
     """
     Find the last known good job for that buildername iterating through the list of revisions.
 
--- a/mozci/scripts/trigger.py
+++ b/mozci/scripts/trigger.py
@@ -111,7 +111,10 @@ def validate_options(options):
 
 
 def sanitize_buildernames(buildernames):
+    """Return the list of buildernames without trailing spaces and with the right capitalization."""
     buildernames_list = buildernames.split(',')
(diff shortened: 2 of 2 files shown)


armenzg:
I like what you did here, however, I would like to see a couple of things being different.

Make sure that all builders in the list are of the same repository.
That way you can just pass one of them to query_repo_url_from_buildername()
Make trigger_range() to accept a list of builders.
You can then iterate over the list of builders here:
https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/mozci.py#L408

I've tested that we can also use -b "builder1,builder2" besides -b "builder1","builder2" (slightly different)

adusca:
I modified trigger_range to receive a list of buildernames, but since backfill_revlist is buildername-specific, we would still need to put almost everything inside the 'for buildername in buildernames' loop (and pass the buildernames to trigger_range one at a time).

I tried modifying backfill_revlist to receive a list of buildernames and return a list of lists of revisions, but that got really messy. A lot of places had to be modified to deal with lists of lists. I believe it is better to not go this way.

Should I keep the modifications in trigger_range? (I did not commit them yet)

armenzg:
I see what you're talking about. I can see how it can get ugly to have different list of revisions for each builder.

Was it what you tried a dictionary like this?

> {
>     'rev1': ['builder1', 'builder2']
>     'rev2': ['builder1']
> }

Anyways, I think your code does the job even though in my mental flow we can tackle builders per revision since they are associated and we fetch the jobs for a revision only once.
However, since we have catching, we're fine to go over all revisions for a different builder.

These comments still apply:
- Make sure that all builders in the list are of the same repository

Could you also take this block and generate a function called determine_revlist()?
https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/scripts/trigger.py#L139-181

It will make it easier to read that for/loop.

Side note, could you please update the comment of trigger_range()? It seems it has fallen out of date.

For my own reference, with the changes in this PR the flow of output would look like this:

> We want to have 1 job(s) of Ubuntu VM 12.04 try opt test mochitest-1 on revisions [u'fbb26646dce6', u'd3297144f8c2', u'74088ab66f44', u'7083568683c9', u'cc9f12c5c7ac']
> === fbb26646dce6 ===
> === d3297144f8c2 ===
> === 74088ab66f44 ===
> === 7083568683c9 ===
> === cc9f12c5c7ac ===
> We want to have 1 job(s) of Ubuntu VM 12.04 try opt test mochitest-2 on revisions [u'fbb26646dce6', u'd3297144f8c2', u'74088ab66f44', u'7083568683c9', u'cc9f12c5c7ac']
> === fbb26646dce6 ===
> === d3297144f8c2 ===
> === 74088ab66f44 ===
> === 7083568683c9 ===
> === cc9f12c5c7ac ===

armenzg:
I saw you commited another 3 changes.
Do you want me to review them?

adusca:
r?

armenzg:
This looks good.
I believe the only thing missing is a doc update for the script.

PR #250, opened 2015-06-08
Refactored query_job_status and added tests for it.

adusca:
The main changes in buildapi.py were:
- moving detecting if a job is coalesced into its on function
- inverting what we call RUNNING with what we call UNKNOWN, as discussed on IRC.

The other changes were just using the fact that we don't need else clauses after return statements to make the code easier to read.


armenzg:
Much cleaner. Thanks!

PR #259, opened 2015-06-15
All coalesced

adusca:
dry-run output: https://pastebin.mozilla.org/8836872


mozci/sources/buildapi.py
```diff
@@ -321,3 +321,14 @@ def query_repositories(clobber=False):
             json.dump(REPOSITORIES, fd)
 
     return REPOSITORIES
+
+
+def find_all_coalesced(repo_name, revision):
```
armenzg: Could you please change this function name to be:
def find_all_by_status(repo_name, revision, status):

That way it becomes more generic.

mozci/sources/buildjson.py
```diff
@@ -141,23 +142,19 @@ def query_job_data(complete_at, request_id):
...
-
-            job = _find_job(request_id, BUILDS_DAY_INDEX[date], filename)
+    job = _find_job(request_id, _fetch_data(filename), filename)
+
+    if job:
+        return job
+
+    # If we have not found the job, it might be that our cache is
+    # old. We will clean the cache and try again
+    global BUILDS_CACHE
```
armenzg: Could you please move this line to after the comments of query_job_data()? I prefer mentioning global variables as soon as we get into a function.

Could you say instead of "try again" the following "try one more time. If this fails, we raise an exception".

armenzg:
Alice, this look very good. Thanks for your prompt help.
There is one major change I would like to see.
Can we please make this part of mozci-trigger? I would like to sheriffs to have a minimal number of entry scripts.

Thank you!

adusca:
Now the usage is:
python trigger.py --coalesced REPO -r REVISION

adusca:
The results of running: python trigger.py --coalesced mozilla-inbound -r ad85daa487
https://treeherder.mozilla.org/#/jobs?repo=mozilla-inbound&revision=ad85daa487cf&filter-resultStatus=success&filter-resultStatus=testfailed&filter-resultStatus=busted&filter-resultStatus=exception&filter-resultStatus=retry&filter-resultStatus=usercancel&filter-resultStatus=running&filter-resultStatus=pending&filter-resultStatus=coalesced

(a screenshot of right after running the script so it is easier to see the jobs triggered by mozci in gray jobs next to the yellow ones. Note that mozci found some coalesced jobs that didn't show up in TH at all, like Linux debug Jit2)
screenshot from 2015-06-15 17 26 06

armenzg:
My apologies ealier.
I did not notice that you used --coallesced to specify the repo name.

Could you please use --repo-name to help keep consistency?

mozci/scripts/trigger.py
```diff
@@ -184,6 +192,16 @@ def main():
         # requests is too noisy and adds no value
         logging.getLogger("requests").setLevel(logging.WARNING)
 
+    if options.coalesced:
+        requests = find_all_by_status(options.coalesced, options.rev, COALESCED)
```
armenzg: Could you call this request_ids instead of requests?
It will be easier to not get it confused with the requests module.

armenzg:
Everything else looks good!

adusca:
No problem. Now the usage is:
python trigger.py --repo-name REPO -r REVISION --coalesced

mozci/scripts/trigger.py
```diff
@@ -86,12 +87,28 @@ def parse_args(argv=None):
...
+                        help="Branch name")
+
     options = parser.parse_args(argv)
     return options
 
 
 def validate_options(options):
     error_message = ""
+    if not options.buildernames and not options.coalesced:
+        error_message = "A buildername is mandatory for all modes except --coalesced."
```
armenzg: Could you also mention to use --buildername?

mozci/scripts/trigger.py
```diff
@@ -86,12 +87,28 @@ def parse_args(argv=None):
...
+
+    if options.coalesced and not options.repo_name:
+        error_message = "A branch name is mandatory with --coalesced."
```
armenzg: Could you also mention to use --repo-name?

armenzg:
Please merge after those two changes.

I'm thinking of having a release today; what do you think?

PR #261, opened 2015-06-16
Continue running in coalesced mode even when there is an exception

adusca:


mozci/sources/buildapi.py
```diff
@@ -330,5 +330,14 @@ def find_all_by_status(repo_name, revision, status):
     Returns a list with the request ids of the coalesced jobs.
     """
     all_jobs = query_jobs_schedule(repo_name, revision)
-    return [job["requests"][0]["request_id"] for job in all_jobs
-            if BuildapiJobStatus(job).get_status() == status]
+    jobs_with_right_status = []
```
vaibhavmagarwal, line 333: Can this be converted to a generic function in mozci.py? Reference: https://github.com/armenzg/mozilla_ci_tools/pull/257/files#r32606864

adusca:
@vaibhavmagarwal Almost every time I tested this script there where some coalesced jobs we were able to find with buildapi that just didn't show up in TH (see the comments in #259 ), so I think it is best to keep this buildapi-only.

mozci/sources/buildapi.py
armenzg, line 333: I think this method should end up under the buildapi class.
Just to be explicit, not every helping method we have for buildapi needs to be implemented for the TH class as well.

mozci/sources/buildjson.py
```diff
@@ -156,7 +156,7 @@ def query_job_data(complete_at, request_id):
     # old. We will clean the cache and try one more time after 60
     # seconds. If it fails, we will raise an Exception
     BUILDS_CACHE = {}
-    LOG.info("The request %d is not yet on %s. We are going to wait for a new %s."
+    LOG.info("The request %d is not yet on %s. We are going to wait 60 seconds for a new %s."
```
armenzg: Could you please define a variable with the value 60 and use that?
I would like sleep to be called like this "time.sleep(seconds)"

That way they will always be kept in sync.

armenzg:
Just the small fix of the sleep variable.
Feel free to merge after that.

We can implement find_all_by_status() on the treeherder class if we need to.

mozci/sources/buildapi.py
vaibhavmagarwal, line 333: Ok, I am glad I asked this, rather than just blindly transferring the function to BuildApi class :)

Issue #242, opened 2015-06-04
Remove --ignore E501 from tox


adusca:
We are currently running flake8 with --ignore E501 [1]. We should remove this flag and fix the warnings this will cause [2]. 

[1] https://github.com/armenzg/mozilla_ci_tools/blob/master/tox.ini#L15

[2] 
mozilla_ci_tools[master $]$ flake8 mozci test
mozci/scripts/alltalos.py:66:101: E501 line too long (103 > 100 characters)
                        help="comma-separated substring filter indicating which jobs not to retrigger")
                                                                                                    ^
mozci/scripts/generate_triggercli.py:64:101: E501 line too long (106 > 100 characters)
        buglist = bugzilla.search_for.summary(options.test_name).keywords("intermittent-failure").search()
                                                                                                    ^
mozci/scripts/misc/find_status_for_jobs.py:66:101: E501 line too long (108 > 100 characters)
                    "&filter-resultStatus=success&filter-resultStatus=testfailed&filter-resultStatus=busted"
                                                                                                    ^
mozci/scripts/misc/find_status_for_jobs.py:67:101: E501 line too long (109 > 100 characters)
                    "&filter-resultStatus=exception&filter-resultStatus=retry&filter-resultStatus=usercancel"
                                                                                                    ^
mozci/scripts/misc/find_status_for_jobs.py:68:101: E501 line too long (109 > 100 characters)
                    "&filter-resultStatus=running&filter-resultStatus=pending&filter-resultStatus=coalesced")
                                                                                                    ^
mozci/utils/transfer.py:134:101: E501 line too long (127 > 100 characters)
            LOG.debug("The local file was last modified in %s. We need to delete the file and fetch it again." % last_mod_date)
                                                                                                    ^
mozci/mozci.py:172:101: E501 line too long (101 > 100 characters)
            LOG.info("We will trigger 1) '%s' instead of 2) '%s'" % (build_buildername, buildername))
                                                                                                    ^
mozci/mozci.py:173:101: E501 line too long (110 > 100 characters)
            LOG.info("We need to trigger the build job once (1) in order to be able to run the test job (2).")
                                                                                                    ^
mozci/mozci.py:364:101: E501 line too long (102 > 100 characters)
                LOG.info("In order to trigger %s %i times, please run the script again after %s ends."
                                                                                                    ^
mozilla_ci_tools[master $]$

priyanklodha:
[...]

PR #264, opened 2015-06-19
Remove --ignore E501 from tox #242

priyanklodha:
What's next?? :D


vaibhavmagarwal:
This isn't on top of the latest master branch. You need to rebase this patch on master.

mozci/scripts/misc/find_status_for_jobs.py
```diff
@@ -63,9 +63,13 @@
...
-                    "&filter-resultStatus=success&filter-resultStatus=testfailed&filter-resultStatus=busted"
-                    "&filter-resultStatus=exception&filter-resultStatus=retry&filter-resultStatus=usercancel"
-                    "&filter-resultStatus=running&filter-resultStatus=pending&filter-resultStatus=coalesced")
+                    "&filter-resultStatus=success&" +
+                    "filter-resultStatus=testfailed&filter-resultStatus=busted"
+
+                    "&filter-resultStatus=exception&" +
+                    "filter-resultStatus=retry&filter-resultStatus=usercancel"
+                    "&filter-resultStatus=running&" +
+                    "filter-resultStatus=pending&filter-resultStatus=coalesced")
```
adusca, line 73: urllib.urlencode({'a': ['b', 'c']}, True) generates a=b&a=c, maybe you could use that here?

mozci/mozci.py
```diff
@@ -169,8 +169,10 @@ def _determine_trigger_objective(revision, buildername, trigger_build_if_missing
             # job multiple times if it is not intentional
             builder_to_trigger = None
         else:
-            LOG.info("We will trigger 1) '%s' instead of 2) '%s'" % (build_buildername, buildername))
-            LOG.info("We need to trigger the build job once (1) in order to be able to run the test job (2).")
+            LOG.info("We will trigger 1) '%s' " +
```
adusca, line 172: Since the string is inside ()s, you don't need to add a + to concatenate the parts, so you can just remove the + sign.

% and + combine in weird ways:

'%s' + '%s' % ('hi, 'hi') is the same as '%s' + ('%s' % ('hi', 'hi')).

Since there is only one %s on the second string, this will raise a TypeError.

('%s' + '%s') % ('hi, 'hi'), would work and '%s' '%s' % ('hi, 'hi') also works.

adusca:
Thanks for the patch! You just need to rebase and make some small changes and we are good to go!

mozci/scripts/generate_triggercli.py
```diff
@@ -61,7 +61,8 @@ def main():
         bugs.append(options.bug_no)
 
     if options.test_name:
-        buglist = bugzilla.search_for.summary(options.test_name).keywords("intermittent-failure").search()
+        buglist = bugzilla.search_for.summary(options.test_name).keywords("intermittent" +
```
adusca, line 64: I think this would be better as:

```
buglist = bugzilla.search_for.summary(
    options.test_name).keywords("intermittent-failure").search()
```

adusca:
I'll close this because it is the same as #266 

@priyanklodha In the future, you can update a PR by updating the branch it came from. The PR gets automatically updated, which makes it easier to track the progress of a patch.

PR #224, opened 2015-05-27
Retrigger existing jobs N times (when possible) instead of triggering N arbitrary jobs

adusca:

PR #265, opened 2015-06-19
Always take the arbitrary build code path when files are directly spe…

chmanchester:
…cified.


adusca:
Thanks @chmanchester! I overlooked this in PR #224, sorry!

PR #266, opened 2015-06-20
E501 issue

priyanklodha:


priyanklodha:
@adusca should i move on some other bug now suggest something?

adusca:
Hi @priyanklodha! Could you please address the other comments I made in https://github.com/armenzg/mozilla_ci_tools/pull/264 ? Thanks!

mozci/scripts/misc/find_status_for_jobs.py
```diff
@@ -57,15 +57,14 @@
...
                     'tochange': revision[0:12],
-                    'filter-searchStr': options.buildername}),
-                    "&filter-resultStatus=success&filter-resultStatus=testfailed&filter-resultStatus=busted"
-                    "&filter-resultStatus=exception&filter-resultStatus=retry&filter-resultStatus=usercancel"
-                    "&filter-resultStatus=running&filter-resultStatus=pending&filter-resultStatus=coalesced")
-
+                    'filter-searchStr': options.buildername,
+                    'filter-resultStatus': ['success', 'testfailed', 'busted'],
+                    'filter-resultStatus': ['exception', 'retry', 'usercancel'],
+                    'filter-resultStatus': ['running', 'pending', 'coalesced']}))
```
adusca: We are almost there. Testing this is a python repl, we get:

```
>>> import urllib
>>> urllib.urlencode({'filter-resultStatus': ['success', 'testfailed', 'busted'],
...                   'filter-resultStatus': ['exception', 'retry', 'usercancel'],
...                   'filter-resultStatus': ['running', 'pending', 'coalesced']})
'filter-resultStatus=%5B%27running%27%2C+%27pending%27%2C+%27coalesced%27%5D'
```

Which is not what we want. 

You need to pass True as the second argument of urlencode in order for lists to work correctly. Also, you must have just one list with every argument of filter-resultStatus. You can test it in the REPL, and you should get the same output as what's currently in the source code (up to ordering)

mozci/scripts/misc/find_status_for_jobs.py
```diff
@@ -63,9 +63,9 @@
...
                     'filter-searchStr': options.buildername}),
-                    "&filter-resultStatus=success&filter-resultStatus=testfailed&filter-resultStatus=busted"
-                    "&filter-resultStatus=exception&filter-resultStatus=retry&filter-resultStatus=usercancel"
-                    "&filter-resultStatus=running&filter-resultStatus=pending&filter-resultStatus=coalesced")
-
+                 urllib.urlencode({
+                     'filter-resultStatus': ['success', 'testfailed', 'busted',
+                                             'exception', 'retry', 'usercancel',
+                                             'running', 'pending', 'coalesced']}, True))
     print "Status of all jobs (success, pending, running, coalesced)"
```
adusca, line 69: This way you get 'https://treeherder.mozilla.org/#/jobs?repo=oi&fromchange=revision&tochange=revision2&filter-searchStr=somethingfilter-resultStatus=success&filter-resultStatus=testfailed&filter-resultStatus=busted&filter-resultStatus=exception&filter-resultStatus=retry&filter-resultStatus=usercancel&filter-resultStatus=running&filter-resultStatus=pending&filter-resultStatus=coalesced'

Note there is no '&' between searchStr=something and filter-resultStatus=success

You don't need a second 'urlib.urlencode. You can just pass 'filter-resultStatus' : ... to the one, remembering to also pass 'True'. 

You will also need to change print 'https://treeherder.mozilla.org/#/jobs?%s%s'  to  print 'https://treeherder.mozilla.org/#/jobs?%s', because now we are only passing one extra string

vaibhavmagarwal:
I really think we could have collapsed this PR in a single commit, rather than 5 separate commits since they basically deal with the same issue.

PR #270, opened 2015-06-29
Added --fill-in mode

adusca:
I asked Ryan what upstream jobs from https://pastebin.mozilla.org/8838040 the sheriffs would like --fill-in mode to trigger. He told me the only one they didn't want was 'mozilla-inbound hg bundle' (and maybe the spider-monkey ones, but he was not sure).

This will trigger every upstream job that is missing from a revision. On non-try repos this will automatically trigger the missing downstream jobs.
Mozilla-inbound output: https://pastebin.mozilla.org/8838045

### DIFF
--- a/mozci/scripts/trigger.py
+++ b/mozci/scripts/trigger.py
@@ -5,6 +5,7 @@ from argparse import ArgumentParser
 
 from mozci.mozci import backfill_revlist, trigger_range, \
     query_repo_name_from_buildername, query_repo_url_from_buildername, query_builders
+from mozci.platforms import get_upstream_buildernames
 from mozci.sources.buildapi import find_all_by_status, make_retrigger_request, COALESCED
 from mozci.sources.pushlog import query_revisions_range_from_revision_and_delta
 from mozci.sources.pushlog import query_revisions_range, query_revision_info, query_pushid_range
@@ -89,6 +90,12 @@ def parse_args(argv=None):
                         help="Trigger every coalesced job on revision --rev "
                         "and repo --repo-name.")
 
+    parser.add_argument("--fill-in",
+                        action="store_true",
+                        dest="fill_in",
+                        help="Fill in missing jobs at revision --rev "
+                        "and repo --repo-name.")
+
     parser.add_argument("--repo-name",
                         dest="repo_name",
                         help="Branch name")
@@ -99,10 +106,10 @@ def parse_args(argv=None):
 
 def validate_options(options):
     error_message = ""
-    if not options.buildernames and not options.coalesced:
-        error_message = "A buildername is mandatory for all modes except --coalesced. " \
-                        "Use --buildername."
-    if options.coalesced and not options.repo_name:
+    if not options.buildernames and not (options.coalesced or options.fill_in):
+        error_message = "A buildername is mandatory for all modes except --coalesced " \
+                        "and --fill-in. Use --buildername."
+    if (options.coalesced or options.fill_in) and not options.repo_name:
         error_message = "A branch name is mandatory with --coalesced. Use --repo-name."
 
     if options.back_revisions:
@@ -205,7 +212,12 @@ def main():
 
         return
 
-    options.buildernames = sanitize_buildernames(options.buildernames)
+    if options.fill_in:
+        options.buildernames = get_upstream_buildernames(options.repo_name)
+
+    else:
+        options.buildernames = sanitize_buildernames(options.buildernames)
+
     repo_url = query_repo_url_from_buildername(options.buildernames[0])
 
     for buildername in options.buildernames:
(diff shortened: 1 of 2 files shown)


mozci/scripts/trigger.py
```diff
@@ -99,10 +106,10 @@ def parse_args(argv=None):
...
     error_message = ""
-    if not options.buildernames and not options.coalesced:
-        error_message = "A buildername is mandatory for all modes except --coalesced. " \
-                        "Use --buildername."
-    if options.coalesced and not options.repo_name:
+    if not options.buildernames and not (options.coalesced or options.fill_in):
+        error_message = "A buildername is mandatory for all modes except --coalesced " \
+                        "and --fill-in. Use --buildername."
+    if (options.coalesced or options.fill_in) and not options.repo_name:
         error_message = "A branch name is mandatory with --coalesced. Use --repo-name."
```
armenzg, line 113: Should you mention in the error message that --repo-name is needed for either --coalesced or --fill-in?

armenzg:
How does it trigger the downstream jobs for non-try repos?
Or what you're saying is that since the build jobs always trigger the downstream jobs we will good? (except if SETA kicks in).

I think we could also approach this another way.
We could look at every possible builder and we ensure that there is at least one job completed for each.
I assume that pulse_actions will have to run again and again for a given revision to ensure that every job is completed (every minute or so).

What do you think? Vaibhav could look into creating a process that goes to sleep for every minute without locking the process. What do you think?

vaibhavmagarwal:
I think what Armen says makes sense. Even if we trigger every upstream job on integration branches, not every downstream job will get triggered after it due to SETA. I think what we want with --fill-in option is to have atleast 1 instance of every job, so we should ideally check for each downstream job instance. This is my interpretation for --fill-in, but you should confirm this with sheriffs.

adusca:
There are 3 types of missing jobs for us to consider:

(1) Missing upstream jobs
(2) Missing downstream jobs for which a finished upstream job already exists
(3) Missing downstream jobs for which no upstream job exits

(1) and (2) can just be triggered directly, the problem is (3). If SETA doesn't kick in just triggering the upstream job is enough, but we cannot count on that.

On the other hand, some build jobs might take a long time to finish (Windows 8 pgo takes ~274 minutes, for example), so keep running the script (even if it is sleeping without locking things) might not be a good idea.

One possible way to deal with that is sending a 'trigger every downstream job when the upstream job finishes' message and let pulse_actions deal with it.

adusca:
The way I suggested for jobs of type (3) before would have security or authentication issues, probably both [1].

I can see 2 ways for the user to run the script locally, both require leaving the script running for a long time:
- Set up a pulse listener to track when the jobs finish: this will require our users to have pulse credentials, but is not very resource hungry and we would be able to trigger jobs as soon as possible
- Periodically check if the jobs we need are finished: this is resource-hungry because we will have to download and process builds-4hr.js several times and there would be some delay between a job finishing and we triggering the downstream jobs.

Or we could make fill-in only a TH button running with pulse_actions and not a mozci-trigger option. 

Or we could add a --fill-in flag that just runs jobs of type (1) and (2) and if users want more jobs they can run the script again after the builds are finished. I don't know what our users would prefer.

[1]
Basically, a malicious user without LDAP credentials could modify the script to send the messages without checking for credentials before and pulse_actions would have no way of knowing the user was not actually authenticated. If special credentials are needed to send pulse messages to the exchange then mozci users would be unable to use the script.

armenzg:
> On the other hand, some build jobs might take a long time to finish
> (Windows 8 pgo takes ~274 minutes, for example), so keep running
> the script (even if it is sleeping without locking things) might not be a good idea.
> 
> One possible way to deal with that is sending a 'trigger every downstream job
> when the upstream job finishes' message and let pulse_actions deal with it.

Very good point. An async approach might be best.

Let's aim for this to be a feature for pulse_actions and not something to run locally. That way we focus on the use case we want to improve.

On a side note, the taskcluster people wrote a cli that does something pretty cool. The script opens a new tab and requires the user to grant permissions to a temporary credential. Once the user accepts in the browser, the user is told that they can close that tab since the script has become authenticated.
It would be cool if our tools could do something like that for pulse, however, I doubt that it would be at all easy with Pulse. In other words, just for your information, you can ask around but I believe there are very complicated technichal difficulties. Perhaps we should file a bug about it and let the TC people add details for one contributor to fix.

armenzg:
I'm moving the converation about pulse_actions here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1178524

adusca:
I think the best option is to make fill-in only a TH button and not a mozci option. @vaibhavmagarwal and @armenzg, do you agree?

armenzg:
Hi Alice,
You're meaning that we should only support this with a pulse listener and forget about also implementing as a CLI function. I agree with this approach since I see it being implemented twice.

vaibhavmagarwal:
I agree that fill-in will mostly (only) be required by sheriffs, so we can just have a TH button

PR #271, opened 2015-07-01
Also consider UNKNOWN jobs as running jobs

adusca:
I forgot to push this a long time ago. Since UNKNOWN means that a job finished running but still isn't successful (and we cannot call query_job_data with it), it should count as running instead of being ignored.


armenzg:
LGTM.
Could you coordinate with @vaibhavmagarwal so his patch or yours don't get bitrotten?

adusca:
This is small, I can wait for @vaibhavmagarwal's patch to land (it should be simple to rebase).

PR #272, opened 2015-07-01
Moved functionality into find_backfill_revlist.

adusca:
I needed this for https://github.com/adusca/pulse_actions/pull/1 and it makes more sense as a mozci function.

In the process of moving this to mozci I noticed that back_revisions, delta and the first part of backfill were all essentially using the same function, so I did some refactoring.

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -459,9 +459,9 @@ def trigger(builder, revision, files=[], dry_run=False, extra_properties=None):
                                           extra_properties)
 
 
-def backfill_revlist(buildername, revisions):
+def _filter_backfill_revlist(buildername, revisions):
     """
-    Find the last known good job for that buildername iterating through the list of revisions.
+    Helper function to find the last known good job for a given buildername on a list of revisions.
 
     If a good job is found, we will only trigger_range() up to that revision instead of the
     complete list (subset of *revlist*).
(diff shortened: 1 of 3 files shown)


mozci/mozci.py
```diff
@@ -453,7 +453,7 @@ def trigger(builder, revision, files=[], dry_run=False, extra_properties=None):
                                           extra_properties)
 
 
-def backfill_revlist(buildername, revisions):
+def filter_backfill_revlist(buildername, revisions):
```
armenzg: Why "filter"?

Maybe find_last_revision_by_matching_status(buildername, revisions, status=SUCCESS) (just a suggestion)

armenzg: Maybe you can add "_" at the beginning of the name to indicate that we should not call this function directly and point to the other one.

mozci/scripts/trigger.py
```diff
@@ -143,17 +143,17 @@ def determine_revlist(repo_url, buildername, rev, back_revisions,
                       delta, from_rev, backfill, skips, max_revisions):
     """Determine which revisions we need to trigger."""
     if back_revisions:
-        push_info = query_revision_info(repo_url, rev)
-        end_id = int(push_info["pushid"])  # newest revision
-        start_id = end_id - back_revisions
-        revlist = query_pushid_range(repo_url=repo_url,
-                                     start_id=start_id,
-                                     end_id=end_id)
+        revlist = query_revisions_range_from_revision_before_and_after(
```
armenzg, line 155: Could you please use named keywords?
repo_url=repo_url, ..., before=..., after=

It was a great recommendation to help read code w/o having to look for the definition of the function.
Think of that zero there. It's hard to tell what it represents without looking.

armenzg:
@adusca is this ready to be reviewed again?

adusca:
r?

PR #295, opened 2015-07-17
Issue 293: Added --existing-only flag to trigger.py

adusca:
r? @vaibhavmagarwal

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -172,6 +172,10 @@ def _determine_trigger_objective(revision, buildername, trigger_build_if_missing
             # This is a safeguard to prevent triggering a build
             # job multiple times if it is not intentional
             builder_to_trigger = None
+            if not trigger_build_if_missing:
+                LOG.info("We would have to trigger build '%s' in order to trigger job '%s'."
+                         " On this mode we will not trigger either." %
+                         (build_buildername, buildername))
         else:
             LOG.info("We will trigger 1) "
                      "'%s' instead of 2) '%s'" % (build_buildername, buildername))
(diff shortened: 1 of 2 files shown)


mozci/scripts/trigger.py
```diff
@@ -98,6 +98,11 @@ def parse_args(argv=None):
                         help="Trigger every coalesced job on revision --rev "
                         "and repo --repo-name.")
 
+    parser.add_argument("--existing-only",
+                        action="store_true",
+                        dest="existing_only",
+                        help="Only trigger test jobs if the build jobs already exists.")
+
```
vaibhavmagarwal, line 105: Can we have the same flag as in all-talos.py: https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/scripts/alltalos.py#L48
This helps bring consistency.

adusca, line 105: The problem is the default is different in both cases.

For alltalos on try the default is not triggering the build, for trigger.py I believe the default should be still be triggering the build. In both cases flags are used to change the default behaviour.

vaibhavmagarwal, line 105: Ok. I think a better way to do this will be:

``` python
parser.add_argument("--existing-only",
                        action="store_false",
                        dest="trigger_build_if_missing",
                        help="Only trigger test jobs if the build jobs already exists.")
```

Doing this will not require this code:

``` python
trigger_build_if_missing = not options.existing_only
```

mozci/mozci.py
```diff
@@ -172,6 +172,10 @@ def _determine_trigger_objective(revision, buildername, trigger_build_if_missing
             # This is a safeguard to prevent triggering a build
             # job multiple times if it is not intentional
             builder_to_trigger = None
+            if not trigger_build_if_missing:
+                LOG.info("We would have to trigger '%s' in order to trigger '%s'."
+                         " On this mode we will not trigger either job."
+                         % (build_buildername, buildername))
         else:
```
vaibhavmagarwal, line 179: Can you add 'build' and 'job' in log.info and put this in 2 lines itself, in this way:

``` python
LOG.info("We would have to trigger build '%s' in order to trigger job '%s'."
         " On this mode we will not trigger either." % (build_buildername, buildername))
```

This makes it clear which is a job and build, to the reader.

vaibhavmagarwal:
r+ with the nits addressed.

adusca:
Thanks, fixed!

PR #297, opened 2015-07-17
Issue 296: Look for 'request_id' key when 'requests' doesn't exist

adusca:
All the jobs I tested had at least one of those keys.

r? @vaibhavmagarwal 

With this patch " python mozci/scripts/triggerbyfilters.py try d81a7a9390a6 -i mochitest-devtools-chrome --times 6 --dry-run" doesn't break.


vaibhavmagarwal:
r+

PR #298, opened 2015-07-21
Add memory-saving mode to buildjson.py

adusca:
This allows pulse_actions to use ijson instead of the normal json module without changing anything about the behaviour of the script.

If the experiment works it might be worth it to make this optimization available for normal mozci usage, but for now I just want to see if this solves our memory problems.

r? @armenzg

### DIFF
--- a/mozci/utils/transfer.py
+++ b/mozci/utils/transfer.py
@@ -1,5 +1,7 @@
 import calendar
+import datetime
 import errno
+import fnmatch
 import gzip
 import json
 import logging
@@ -8,14 +10,20 @@ import platform
 import shutil
 import subprocess
 import time
-import datetime
-import fnmatch
 
 import requests
 
 from progressbar import Bar, Timer, FileTransferSpeed, ProgressBar
 
+# yajl2 backend is faster then the default backend, but it requires
+# libyajl2 to be installed in the system
+try:
+    import ijson.backends.yajl2 as ijson
+except:
+    import ijson
+
 LOG = logging.getLogger('mozci')
+MEMORY_SAVING_MODE = False
 
 
 def path_to_file(filename):
@@ -186,10 +194,38 @@ def load_file(filename, url):
         raise Exception("We received %s which is unexpected." % req.status_code)
 
     try:
-        return _load_json_file(filepath)
+        if not MEMORY_SAVING_MODE:
+            return _load_json_file(filepath)
+
+        return _lean_load_json_file(filepath)
 
     # Issue 213: sometimes we download a corrupted builds-*.js file
     except IOError:
         LOG.info("%s is corrupted, we will have to download a new one.", filename)
         os.remove(filepath)
         return load_file(filename, url)
+
+
+def _lean_load_json_file(filepath):
+    """Helper function to load json contents from a file using ijson."""
+    LOG.debug("About to load %s." % filepath)
+
+    fd = open(filepath, 'r')
+
+    gzipper = gzip.GzipFile(fileobj=fd)
+    builds = ijson.items(gzipper, 'builds.item')
+    ret = {'builds': []}
+    # We are going to store only the information we need from builds-.js
+    # and drop the rest.
+    ret['builds'] = [
+        {"properties": {
+            "buildername": b["properties"].get("buildername", None),
+            "request_ids": b["properties"].get("request_ids", []),
+            "revision": b["properties"].get("revision", None)},
+         "request_ids": b["request_ids"]}
+        for b in builds]
+
+    fd.close()
+    gzipper.close()
+
+    return ret
--- a/setup.py
+++ b/setup.py
@@ -14,6 +14,7 @@ setup(
     install_requires=[
         'beautifulsoup4>=4.3.2',
         'bugsy>=0.4.0',
+        'ijson>=2.2',
         'progressbar>=2.3',
         'requests>=2.5.1',
         'keyring>=5.3',


adusca:
Memory profiling results running: 

```
python -m memory_profiler mozci/scripts/trigger.py --dry-run --coalesced --repo-name mozilla-inbound -r  649648c5f257
```

| LEAN | Maximum memory usage |
| --- | --- |
| False | 348.145 MiB |
| True | 58.480 MiB |

PS.: For some reason, the script with LEAN=True is a lot slower with memory_profiler, even though it has a similar speed with normal usage when libyajl2.c is installed in the system.

mozci/utils/transfer.py
```diff
@@ -8,13 +10,18 @@
...
 import time
-import datetime
-import fnmatch
 
 import requests
 
 from progressbar import Bar, Timer, FileTransferSpeed, ProgressBar
 
+# yajl2 backend is faster then the default backend, but it requires
+# libyajl2 to be installed in the system
```
vaibhavmagarwal, line 19: Just a drive-by. Is libyajl2 installed in the system with ijson? (since I do not see it included in setup.py)

adusca, line 19: Unfortunately, it is a C dependency which cannot be easily installed with pip, and it doesn't get automatically installed with ijson. Some projects, like [0], have commands in setup.py to manually compile and install libyajl2, but I'm hoping we will be able to configure Heroku so that libyajl2 gets installed through a package manager (Debian/Ubuntu have a libyajl2 package, for example). 

I tried a bit and couldn't get it to run in Travis [1,2], including installing other YAJL related packages which could try to install the C library themselves, but we can do something like [0] if it comes to that.

[0] https://github.com/rtyler/py-yajl/blob/master/setup.py
[1] https://travis-ci.org/armenzg/mozilla_ci_tools/pull_requests
[2] https://travis-ci.org/armenzg/mozilla_ci_tools/builds/71867726

mozci/sources/buildjson.py
```diff
@@ -14,7 +14,7 @@
 BUILDJSON_DATA = "http://builddata.pub.build.mozilla.org/builddata/buildjson"
 BUILDS_4HR_FILE = "builds-4hr.js"
 BUILDS_DAY_FILE = "builds-%s.js"
-
+LEAN = False
 # This helps us read into memory and load less from disk
```
vaibhavmagarwal, line 18: Instead of passing "lean" parameters between functions, you could define it in transfer.py as a global variable. And do:

``` python
from mozci.utils.transfer import LEAN
```

You could then use it as a global variable and specify its value in one place rather than as arguments, IMO, that makes it easier. Also, personally I prefer more descriptive names, does "MEMORY_SAVING_MODE" make more sense than LEAN to you?

armenzg, line 18: I agree with your suggetion.
I was about to ask what LEAN means!

mozci/utils/transfer.py
```diff
@@ -186,10 +193,37 @@ def load_file(filename, url):
...
+    gzipper = gzip.GzipFile(fileobj=fd)
+    builds = ijson.items(gzipper, 'builds.item')
+    ret = {'builds': []}
+
+    ret['builds'] = [
+        {"properties": {
+            "buildername": b["properties"].get("buildername", None),
+            "request_ids": b["properties"].get("request_ids", []),
+            "revision": b["properties"].get("revision", None)},
+         "request_ids": b["request_ids"]}
```
armenzg, line 225: When you iterate through the data, do you drop the data that we don't need from the json file?
I think that is what you're doing but a comment would help anyone reading this code.

armenzg:
Alice, when you use this on Heroku, how will you be telling it to use the memory saving mode?

This looks good. I hope it easily works. It can always be funky to run code in infra you don't completely own.

I see Travis passing. What did you do differently?

On a side note (different PR), I think we need to look into clear the memory where possible.
I found this answer on stackoverflow interesting:
http://stackoverflow.com/a/11058050/513672
and what seems a good video about memory usage and Python:
http://pyvideo.org/video/717/python-linkers-and-virtual-memory

adusca:
> when you use this on Heroku, how will you be telling it to use the memory saving mode?

My plan was to first deal with this by setting the variable to True on pulse actions to see if it works and if it works think of a more robust solution, but I know this is ugly and I'm open to suggestions :)

> I see Travis passing. What did you do differently?Alice, when you use this on Heroku, how will you be telling it to use the memory saving mode?

ijson has 2 possible backends: yajl (yet-another-json-library), that depends on a C library and one backend in pure Python. The yajl one is much faster (on my system it was around 10x faster), but it will raise import errors on systems that don't have the library installed. I added the 'try' block so when that happens ijson falls back to the pure Python backend.

adusca:
I updated the patch to address the comments. r? @armenzg

mozci/sources/buildjson.py
```diff
@@ -14,7 +14,6 @@
 BUILDJSON_DATA = "http://builddata.pub.build.mozilla.org/builddata/buildjson"
 BUILDS_4HR_FILE = "builds-4hr.js"
 BUILDS_DAY_FILE = "builds-%s.js"
-
 # This helps us read into memory and load less from disk
```
vaibhavmagarwal, line 18: Could you add the blank line back?

adusca:
Blank line added.

armenzg:
This looks very nice. Please go ahead and merge it if you're satisfied.

Let me know how the deployment in pulse_actions goes.

Issue #299, opened 2015-07-22
mozci is busted on Windows


rvandermeulen:
$ mozci-trigger -b "Android armv7 API 9 mozilla-central debug build" -r e7434cafdf2f
07/22/2015 10:57:20 INFO:        We want to have 1 job(s) of Android armv7 API 9 mozilla-central deb
ug build on revisions ['e7434cafdf2f']
07/22/2015 10:57:20 INFO:
07/22/2015 10:57:20 INFO:        === e7434cafdf2f ===
07/22/2015 10:57:21 INFO:        We want to have 1 job(s) of Android armv7 API 9 mozilla-central deb
ug build on revision e7434cafdf2f
builds-4hr.js: [>>>>>>>>>>>>>>>>>>>>>>] Elapsed Time: 0:00:01   2.72 MB/s 3.0MB

gzip: c:/Users/Ryan/.mozilla/mozci/builds-4hr.js: invalid compressed data--format violated
07/22/2015 10:57:23 ERROR:       'CalledProcessError' object has no attribute 'errno'
Traceback (most recent call last):
  File "c:\mozilla-build\python\lib\site-packages\mozci\scripts\trigger.py", line 251, in main
    trigger_build_if_missing=options.trigger_build_if_missing
  File "c:\mozilla-build\python\lib\site-packages\mozci\mozci.py", line 387, in trigger_range
    successful_jobs, pending_jobs, running_jobs = _status_summary(matching_jobs)[0:3]
  File "c:\mozilla-build\python\lib\site-packages\mozci\mozci.py", line 69, in _status_summary
    status = QUERY_SOURCE.get_job_status(job)
  File "c:\mozilla-build\python\lib\site-packages\mozci\query_jobs.py", line 92, in get_job_status
    return self._is_coalesced(job)
  File "c:\mozilla-build\python\lib\site-packages\mozci\query_jobs.py", line 104, in _is_coalesced
    status_data = query_job_data(req["complete_at"], req["request_id"])
  File "c:\mozilla-build\python\lib\site-packages\mozci\sources\buildjson.py", line 147, in query_jo
b_data
    job = _find_job(request_id, _fetch_data(filename), filename)
  File "c:\mozilla-build\python\lib\site-packages\mozci\sources\buildjson.py", line 45, in _fetch_da
ta
    json_contents = load_file(filepath, url)
  File "c:\mozilla-build\python\lib\site-packages\mozci\utils\transfer.py", line 189, in load_file
[...]

Issue #213, opened 2015-05-19
IOError: CRC check failed - happens quite often (5 times today already)


jmaher:
Traceback (most recent call last):
  File "triggerbyfilters.py", line 108, in <module>
    main()
  File "triggerbyfilters.py", line 97, in main
    dry_run=options.dry_run,
  File "/home/jmaher/mozilla/mozilla_ci_tools/mozci/mozci.py", line 415, in trigger_range
    files=files)
  File "/home/jmaher/mozilla/mozilla_ci_tools/mozci/mozci.py", line 344, in trigger_job
    buildername,
  File "/home/jmaher/mozilla/mozilla_ci_tools/mozci/mozci.py", line 132, in _determine_trigger_objective
    files = _find_files(job)
  File "/home/jmaher/mozilla/mozilla_ci_tools/mozci/mozci.py", line 188, in _find_files
    job_status = _status_info(job_schedule_info)
  File "/home/jmaher/mozilla/mozilla_ci_tools/mozci/mozci.py", line 181, in _status_info
    return buildjson.query_job_data(complete_at, request_id)
  File "/home/jmaher/mozilla/mozilla_ci_tools/mozci/sources/buildjson.py", line 146, in query_job_data
    job = _find_job(request_id, _fetch_data(BUILDS_4HR_FILE), BUILDS_4HR_FILE)
  File "/home/jmaher/mozilla/mozilla_ci_tools/mozci/sources/buildjson.py", line 42, in _fetch_data
    json_contents = load_file(filepath, url)
  File "/home/jmaher/mozilla/mozilla_ci_tools/mozci/utils/transfer.py", line 156, in load_file
    return _fetch_and_load_file(req, filepath)
  File "/home/jmaher/mozilla/mozilla_ci_tools/mozci/utils/transfer.py", line 87, in _fetch_and_load_file
    blob = gzipper.read()
  File "/usr/lib/python2.7/gzip.py", line 254, in read
    self._read(readsize)
  File "/usr/lib/python2.7/gzip.py", line 308, in _read
    self._read_eof()
  File "/usr/lib/python2.7/gzip.py", line 347, in _read_eof
    hex(self.crc)))
IOError: CRC check failed 0x3d16ffd8 != 0x9fa335b8L

jmaher:
and the magical commandline:
python triggerbyfilters.py --times 2 -i mochitest -e devtools,chrome,other,android,b2g,pgo,10.8,Code try e3c7e03823c5

[...]

PR #300, opened 2015-07-22
Issue 299: Dealing better with corrupted files on Windows.

adusca:
Turns out #299 is the Windows version of #213. The error raised for corrupted files is different on Windows (it raises subprocess.CalledProcessError instead of IOError). Fixing this revealed another bug: the file was corrupted 100% of the time because it was opened in text mode when writing and reading, which converts \n bytes to \r\n on Windows.

The error message was so weird because there was another bug in the code: we were assuming that we would get a subprocess.CalledProcessError when gzip is not installed on the system but we actually get a OSError.

1 issue, 3 bugs. Windows is weird :)

### DIFF
--- a/mozci/utils/transfer.py
+++ b/mozci/utils/transfer.py
@@ -86,7 +86,7 @@ def _load_json_file(filepath):
     LOG.debug("About to load %s." % filepath)
 
     # Sniff whether the file is gzipped
-    fd = open(filepath, 'r')
+    fd = open(filepath, 'rb')
     magic = fd.read(2)
     fd.seek(0)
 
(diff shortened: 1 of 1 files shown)


vaibhavmagarwal:
r+

vaibhavmagarwal:
@adusca thanks for the patch. Does this fully resolve https://github.com/armenzg/mozilla_ci_tools/issues/299 ? And does it also resolve some other issue (since you say 3 bugs are fixed with this)?

armenzg:
Cool! Thank you both.
FYI no need to do a release for this since it was more of a pass-by by Ryan. We don't have enough Windows users to rush a relase out the door.

PR #302, opened 2015-07-23
Issue 301 - Switch to new TravisCI infrastructure

armenzg:
By switching to the new insfrastructure we get:
- more isolation (thanks to containers)
- faster builds
- CPU resources are guaranteed
- caching available

caveat: As you can’t use sudo on the new container-based infrastructure,
you need to use the addons.apt.packages and addons.apt.sources plugins
to install packages and package sources

[1]
http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+sudo: false
+
 language: python
 
 python:


armenzg:
r?

I will have to wait on Travis results before merging.

vaibhavmagarwal:
r+, feel free to merge in

armenzg:
The faster Travis CI infras was not substantially faster to execute but maybe the wait time for an available machine was shorter

It adds the output to the build:

> This job is running on container-based infrastructure, which does not allow use of 'sudo', setuid and setguid executables.
> If you require sudo, add 'sudo: required' to your .travis.yml
> See http://docs.travis-ci.com/user/workers/container-based-infrastructure/ for details.

PR #305, opened 2015-07-29
Also count failed jobs as potential jobs

adusca:
Today, during the automatic backfilling experiment, we noticed that sometimes mozci would trigger a job after calling trigger_range with times=1 when 5 runs of the job already existed.

After investigating, I discovered that this is because mozci didn't consider failed jobs as potential jobs. I know that for pulse_actions we definitely don't want to trigger a job extra times because all of the previous runs were failures, but I'm not sure we should add an argument to keep the previous behaviour or if counting failed jobs as potential jobs is something we always want to do in mozci.

r? @armenzg

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -17,7 +17,7 @@ import logging
 from mozci.platforms import determine_upstream_builder, is_downstream, filter_buildernames
 from mozci.sources import allthethings, buildapi, buildjson, pushlog
 from mozci.query_jobs import PENDING, RUNNING, SUCCESS, UNKNOWN,\
-    COALESCED, BuildApi, TreeherderApi
+    COALESCED, WARNING, FAILURE, EXCEPTION, RETRY, BuildApi, TreeherderApi
 from mozci.utils.misc import _all_urls_reachable
 from mozci.utils.transfer import path_to_file, clean_directory
 
@@ -64,21 +64,22 @@ def _status_summary(jobs):
     pending = 0
     running = 0
     coalesced = 0
+    failed = 0
 
     for job in jobs:
         status = QUERY_SOURCE.get_job_status(job)
         if status == PENDING:
             pending += 1
-        if status == RUNNING:
+        if status in (RUNNING, UNKNOWN):
             running += 1
         if status == SUCCESS:
             successful += 1
         if status == COALESCED:
             coalesced += 1
-        if status == UNKNOWN:
-            running += 1
+        if status in (FAILURE, WARNING, EXCEPTION, RETRY):
+            failed += 1
 
-    return (successful, pending, running, coalesced)
+    return (successful, pending, running, coalesced, failed)
 
 
 def _determine_trigger_objective(revision, buildername, trigger_build_if_missing=True):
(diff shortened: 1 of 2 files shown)


mozci/mozci.py
```diff
@@ -64,21 +64,22 @@ def _status_summary(jobs):
     pending = 0
     running = 0
     coalesced = 0
+    failed = 0
```
armenzg, line 67: I'm going to file a re-factoring issue for this function since this is very hard-coded.

mozci/mozci.py
```diff
@@ -64,21 +64,22 @@ def _status_summary(jobs):
...
 
     for job in jobs:
         status = QUERY_SOURCE.get_job_status(job)
         if status == PENDING:
             pending += 1
-        if status == RUNNING:
+        if status == RUNNING or status == UNKNOWN:
```
armenzg: Could you please change this to status in (RUNNING, UNKNOWN)? Same below.

mozci/mozci.py
```diff
@@ -384,11 +385,11 @@ def trigger_range(buildername, revisions, times=1, dry_run=False,
 
         # 1) How many potentially completed jobs can we get for this buildername?
         matching_jobs = QUERY_SOURCE.get_matching_jobs(repo_name, rev, buildername)
-        successful_jobs, pending_jobs, running_jobs = _status_summary(matching_jobs)[0:3]
+        successful_jobs, pending_jobs, running_jobs, _, failed_jobs = _status_summary(matching_jobs)
 
-        potential_jobs = pending_jobs + running_jobs + successful_jobs
-        LOG.debug("We found %d pending jobs, %d running jobs and %d successful_jobs." %
-                  (pending_jobs, running_jobs, successful_jobs))
+        potential_jobs = pending_jobs + running_jobs + successful_jobs + failed_jobs
```
armenzg, line 390: Including failed_jobs into potential_jobs will affect the behaviour for all other use cases where the intent is to obtain an exact number of jobs in a specific state (in the current code we intent to obtain a minimum number of successful jobs; pending/running are the real _potentially_ successful jobs - meaning that I should have not mentioned successful in here.)

Perhaps, we should add an extra parameter which we set as a default for it.
valid_statuses = (PENDING, RUNNING, SUCCESS)

We could add a variable called "wanted_statuses" which would be jobs matching valid_statuses. potential_jobs would only contain pending and running.

Depends on how fast you would like us to go, I will be happy to merge this in as-is but I would like us to re-factor this.
I will mention my re-factoring comments and you tell me what you prefer.

mozci/mozci.py
```diff
@@ -384,11 +385,11 @@ def trigger_range(buildername, revisions, times=1, dry_run=False,
...
+        LOG.debug("We found %d pending jobs, %d running jobs, %d successful_jobs and "
+                  "%d failed jobs" % (pending_jobs, running_jobs, successful_jobs, failed_jobs))
```
armenzg: We probably should change this to:
"We have P/R pending/running jobs, X jobs in {list of valid statuses} and Y jobs in remaining statuses".

mozci/mozci.py
vaibhavmagarwal, line 390: In my opinion "FAILED" jobs are also jobs that are run, and they should be counted in total number of jobs.

armenzg, line 390: Yes, but up until now, we've tried to ensure an N number of successful jobs.
We have to allow each user of trigger_range() what should be considered.
For instance, for automated backfilling we want 1 completed job. For the case of talos, we might want 5 successful jobs.

Adding more flexibility is good but we might want to still make sure that we add a security measure.
What about if we add a parameter being max_completed_jobs=None?
If no value is set, we set it to times?
For max_completed_jobs we consider all completed jobs regardless of status.

By doing this:
- we can say we want 1 green job but do not try more than 5 times.
- we also prevent triggering forever.

adusca:
I made the changes that were small. I think I prefer to leave _status_summary refactoring as a separate issue (I added a TODO to remember to change the debug comment after the refactoring).

This patch doesn't affect alltalos.py or --coalesced mode, because they don't use trigger_range. It does affect every other mode of trigger.py, but I'm hoping our users will prefer the new behaviour. If they don't, we can go back to the previous behaviour and add a parameter for total jobs.

armenzg:
WFM all yours!

armenzg:
Alice, would you be OK if we did this?
potential_jobs = pending_jobs + running_jobs + completed_jobs

That way it doesn't matter if the jobs are red, orange, blue or purple.
I'm afraid we might hit other cases of permanent state (in our original case it was a warning state).

mozci/mozci.py
```diff
@@ -64,21 +64,22 @@ def _status_summary(jobs):
...
-        if status == RUNNING:
+        if status in (RUNNING, UNKNOWN):
             running += 1
         if status == SUCCESS:
             successful += 1
         if status == COALESCED:
             coalesced += 1
-        if status == UNKNOWN:
-            running += 1
+        if status in (FAILURE, WARNING, EXCEPTION, RETRY):
```
adusca, line 79: Possible status are:
PENDING, RUNNING, COALESCED, UNKNOWN
SUCCESS, WARNING, FAILURE, SKIPPED, EXCEPTION, RETRY, CANCELLED

Of those I think the complete-but-not-successful ones are FAILURE, WARNING, EXCEPTION, RETRY.

This way potential_jobs = pending_jobs + running_jobs + completed_jobs, but we still get some info on how many jobs were successful.

PR #308, opened 2015-07-31
Bug 1149738 - Add the ability to retrigger TaskCluster tasks

armenzg:
This code requires setting up a couple of env variables that are the credentials for a client in TaskCluster
This also solves a logging issue where we would only see the logging of modules named 'mozci'

r? @chmanchester


mozci/scheduling.py
```diff
@@ -0,0 +1,28 @@
...
+in a very generic manner.
+
+Defined in here:
+    * BaseSchedulingClient
+    * TaskclusterSchedulingClient
+"""
+from __future__ import absolute_import
+from abc import ABCMeta, abstractmethod
+
+from mozci.sources import taskcluster_
```
gregarndt: why the underscore?

armenzg: IIRC it was because the taskcluster client is named "taskcluster" so there is a bit of conflict. I didn't look further to see how to work around it.

armenzg: @gregarndt did the rest look OK?

mozci/sources/taskcluster_.py
```diff
@@ -0,0 +1,86 @@
...
+LOG = logging.getLogger('mozci')
+TASKCLUSTER_TOOLS_HOST = 'https://tools.taskcluster.net'
+
+
+def retrigger_task(task_id, dry_run=False):
+    """ Given a task id (our uuid) we  query it and build
+    a new task based on the old one which we schedule on TaskCluster.
+
+    We don't call the rerun API since we can't rerun a task past
+    its deadline, instead we create a new task with a new taskGroupId,
```
gregarndt: So you use a new task group id...do you not want it to appear in the same graph as the other tasks?

mozci/sources/taskcluster_.py
```diff
@@ -0,0 +1,86 @@
...
+    http://docs.taskcluster.net/queue/api-docs/#createTask
+    """
+    one_year = 365
+    new_task_id = None
+
+    try:
+        queue = taskcluster_client.Queue()
+        task = queue.task(task_id)
+
+        LOG.debug("Original task: (Limit 1024 char)")
```
gregarndt: Just curious, why the limit here?

mozci/sources/taskcluster_.py
armenzg: No technical reason. I re-used the code you pasted in one of the bugs.
I thought there was a reason to not touch a graph once all existing tasks had completed or expired.

mozci/sources/taskcluster_.py
armenzg: The debug output can be rather noisy.
I saw this approach used for the taskcluster client.
I assume that if anyone would like to see more they could modify the code while they're debugging an issue (since they're using debug output to pin-point something).

I'm not strongly attached to this. It simply made my life easier when using debug output.

gregarndt: Thanks for the explanation, was just curious.  I have not look looked much at the client code.

mozci/sources/taskcluster_.py
gregarndt: Ah, you're right. Looking more into the docs, we should not be retriggering tasks that are part of a graph that has been marked as blocked or completed.  Modifying the task graph should only happen if the task graph is still in an open state.

mozci/sources/taskcluster_.py
```diff
@@ -0,0 +1,86 @@
...
+        artifacts = task['payload'].get('artifacts', {})
+        for artifact, definition in artifacts.iteritems():
+            definition['expires'] = taskcluster_client.fromNow('%s days' % one_year)
+
+        # The task group will be identified by the ID of the only
+        # task in the group
+        task['taskGroupId'] = new_task_id
+        # TC workers create public logs which are 365 days; if the task expiration
+        # date is the same or less than that we won't have logs for the task
+        task['expires'] = taskcluster_client.fromNow('%s days' % (one_year + 1))
```
gregarndt: Not sure if you want to track bugs related to the reasoning here, but we're looking at fixing the worker to make sure the expiration of the log artifact is not beyond the task.expires.

https://bugzilla.mozilla.org/show_bug.cgi?id=1190660

mozci/sources/taskcluster_.py
```diff
@@ -0,0 +1,86 @@
...
+        task['created'] = taskcluster_client.stringDate(datetime.datetime.utcnow())
+        task['deadline'] = taskcluster_client.fromNow('24 hours')
```
gregarndt: this is completely just a "cool" thing, but not sure if you could calculate the original delta between created and deadline, and set the new deadline based off of that.

mozci/sources/taskcluster_.py
```diff
@@ -0,0 +1,86 @@
...
+        LOG.debug("Contents of new task: (Limit 1024 char)")
+        LOG.debug(str(task)[:1024])
+
+        if not dry_run:
+            LOG.info("Attempting to schedule new task with task_id: {}".format(new_task_id))
+            result = queue.createTask(new_task_id, task)
+            LOG.debug(result)
+            LOG.info("{}/task-inspector/#{}".format(TASKCLUSTER_TOOLS_HOST, new_task_id))
+        else:
+            LOG.info("Dry-run mode: Nothing was retriggered.")
```
gregarndt: Would outputting the entire task as JSON be valuable here for dry runs?

mozci/scripts/misc/taskcluster_retrigger.py
```diff
@@ -0,0 +1,49 @@
...
+
+    if options.debug:
+        setup_logging(logging.DEBUG)
+    else:
+        setup_logging()
+
+    if options.retrigger:
+        sch = TaskclusterSchedulingClient()
+        for t_id in options.task_ids:
+            sch.retrigger(uuid=t_id, dry_run=options.dry_run)
```
gregarndt: so it looks like the new task id is returned and we just exit successfully even if there is an authorization/authentication failure and the new task id is -1.  Is there value in existing with a non-zero exit code if the task could not be scheduled?

armenzg: As you point it out, at this moment we don't really do anything with a bad outcome.
Let's see what Chris would prefer since he is going to be the first consumer of this functionality.

mozci/mozci.py
```diff
@@ -14,10 +14,10 @@
 
 import logging
 
-from mozci.platforms import determine_upstream_builder, is_downstream, filter_buildernames,\
-    build_talos_buildernames_for_repo
+from mozci.platforms import determine_upstream_builder, is_downstream, \
+    filter_buildernames, build_talos_buildernames_for_repo
 from mozci.sources import allthethings, buildapi, buildjson, pushlog
-from mozci.query_jobs import PENDING, RUNNING, SUCCESS, UNKNOWN,\
+from mozci.query_jobs import PENDING, RUNNING, SUCCESS, UNKNOWN, \
```
chmanchester: While we're here, a style I see in tree a lot is 

``` python
from mozci.query import (
    PENDING,
    RUNNING,
    ...
)
```

for when the import block would span multiple lines.

mozci/scheduling.py
chmanchester: Using a fully qualified import might be a good way around this.

mozci/scheduling.py
```diff
@@ -0,0 +1,28 @@
+"""
+This module allow us to interact with the various scheduling systems
+in a very generic manner.
+
+Defined in here:
+    * BaseSchedulingClient
+    * TaskclusterSchedulingClient
+"""
```
chmanchester, line 8: This module doesn't seem to live up to its stated purpose here -- the only concrete implementation is for TaskCluster. Last we spoke about this the issue was that TaskCluster doesn't store task ids per revision/tree for us, so we couldn't have an interface like the one we have for buildapi. Do we have plans for this transition beyond what's here?

mozci/scripts/misc/taskcluster_retrigger.py
```diff
@@ -0,0 +1,49 @@
...
+from mozci.scheduling import TaskclusterSchedulingClient
+from mozci.utils.misc import setup_logging
+
+
+def main():
+    parser = ArgumentParser()
+    parser.add_argument('-r',
+                        action="store_true",
+                        dest="retrigger",
+                        help="It retriggers a TaskCluster task.")
```
chmanchester: The script doesn't seem to do anything when this isn't present, can we leave out the argument?

mozci/scripts/misc/taskcluster_retrigger.py
chmanchester: I'm not planning to use this from the command line -- the behavior here isn't crucial (although if I were, I would expect a message logged with the task ids that failed to get retriggered).

mozci/scheduling.py
```diff
@@ -0,0 +1,28 @@
...
+
+    @abstractmethod
+    def retrigger(self, uuid, **kwargs):
+        pass
+
+
+class TaskclusterSchedulingClient(BaseSchedulingClient):
+
+    def retrigger(self, uuid, **kwargs):
+        taskcluster_.retrigger_task(task_id=uuid, **kwargs)
```
chmanchester: We might want the new task id returned here.

mozci/scheduling.py
armenzg, line 8: Right now, I only want to add the minimal parts for TC rather than also add the pieces for BuildAPI to make this PR more focused.

I can state it as such.

I'm hoping that in the long term we would write a BuildapiSchedulingClient which would use the buildapi client (mozci/sources/buildapi.py)

chmanchester, line 8: Ok, I guess treeherder must correlate Task IDs with pushes and trees, I'll check there.

mozci/utils/misc.py
```diff
@@ -53,18 +53,21 @@ def setup_logging(level):
     As seen in:
     https://docs.python.org/2/howto/logging-cookbook.html#logging-to-multiple-destinations
     """
-    LOG = logging.getLogger('mozci')
+    LOG = logging.getLogger()
```
chmanchester, line 56: I should have noticed this before, but why stop getting the logger by name? In tc.py we still have `logging.getLogger('mozci')`

armenzg, line 56: If I recall correctly, the reason is that if we set "mozci" as the name, we then loose all logging messages from other modules like the "taskcluster.client" or even "requests".

In fact, this is a fix of a current issue where we're loosing debugging abilities when we call anything with --debug.

chmanchester, line 56: I see now, you're configuring the root logger. Makes sense, thanks for the explanation!

Issue #313, opened 2015-08-07
Add ability to trigger new builds if revision not found in buildapi


vaibhavmagarwal:
RyanVM said that he mistakenly pushed with 'DONT BUILD' to treeherder, and tried to trigger missing jobs. However mozci doesn't do anything if it can't find the particular revision. We should trigger new builds in such a case, since we have the ability to trigger_arbitrary_jobs. Relevant logs:
```python
INFO:	 We want to have 1 job(s) of Android 4.0 armv7 API 11+ mozilla-central debug test crashtest on revisions [u'be100e46b1ba'] 
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	  
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	 === be100e46b1ba === 
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	 We can't trigger anything on pushes that the revision is not valid for buildapi. 
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	 We want to have 1 job(s) of Rev4 MacOSX Snow Leopard 10.6 mozilla-central opt test mochitest-2 on revisions [u'be100e46b1ba'] 
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	  
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	 === be100e46b1ba === 
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	 We can't trigger anything on pushes that the revision is not valid for buildapi. 
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	 We want to have 1 job(s) of Rev4 MacOSX Snow Leopard 10.6 mozilla-central opt test mochitest-3 on revisions [u'be100e46b1ba'] 
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	  
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	 === be100e46b1ba === 
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	 We can't trigger anything on pushes that the revision is not valid for buildapi. 
Aug 07 12:48:52 pulse-actions app/worker2.1: INFO:	 We want to have 1 job(s) of Rev4 MacOSX Snow Leopard 10.6 mozilla-central opt test mochitest-4 on revisions [u'be100e46b1ba'] 
[...]

PR #315, opened 2015-08-08
Using pushlog to validate revisions

adusca:
I still need to update the tests, but this is the general idea. Since we discovered that a revision doesn't have to exist on buildapi for trigger_arbitrary_job to work (see comments in #313 ).

This allows us to trigger stuff in revisions on try with only TC jobs and in DONTBUILD revisions.


mozci/sources/buildapi.py
```diff
@@ -157,49 +157,23 @@ def _valid_builder():
                     "mozci.allthethings.")
 
 
-def valid_revision(repo_name, revision):
```
armenzg, line 160: We might still want to keep this around (but with a different name) to validate credentials of a user as soon as possible.

mozci/scripts/triggerbyfilters.py
```diff
@@ -75,7 +76,8 @@ def main():
         LOG = setup_logging(logging.INFO)
 
     if options.rev == 'tip':
-        options.rev = query_repo_tip(options.repo)
+        repo_url = query_repo_url(options.repo)
+        options.rev = query_repo_tip(repo_url)
```
armenzg, line 81: The day we decide to move pushlog as its own client we might want to teach it how to determine the repo_url by itself.
Thank you for removing the dependency on buildapi.

mozci/sources/buildapi.py
armenzg, line 160: More clearly, rename the function to validate credentials and use another end-point which will always be available (unless this one with a revision).

armenzg:
All looking good.
I need that removed function to be addressed.
Merge once satisfied. Good job!

mozci/sources/buildapi.py
adusca, line 160: I added a function called valid_crendentials that calls get_credentials and raises an AuthenticationError when the credentials are invalid, but I'm not sure where we should call this function.

adusca, line 160: I'm calling valid_credentials as the first step in both scripts, so we ask for credentials as soon as possible.

armenzg:
Beautiful :+1:

PR #325, opened 2015-08-14
Using @patch in TestQueries.

adusca:
Writing #315 I had testing problems caused by improper mocking. This will make our tests more robust and easier to debug.

r?


vaibhavmagarwal:
r+

PR #329, opened 2015-08-17
Issue 328: Do not prompt for password too soon

adusca:


armenzg:
One more bites the dust!

PR #337, opened 2015-08-20
Allow Mozci to schedule buildbot jobs through TaskCluster

armenzg:
You can run something like this:

> python mozci/scripts/misc/buildbot_to_taskcluster.py --repo-name try
> '{"Android armv7 API 9 try build": ["Android armv7 API 9 try opt test plain-reftest-16"]}' 
> --revision da280e3f1ab3 | taskcluster run-graph --verbose

@adusca would you mind reviewing this PR?

In my next PR I will remove the need to pipe it to the taskcluster binary and add dry_run mode.


mozci/sources/buildbot_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+
+from mozci.sources.buildapi import query_repo_url
+from mozci.sources.pushlog import query_revision_info
+from mozci.platforms import get_builder_information
+
+
+def _create_task(buildername, repo_name, revision, metadata, requires=None):
+    """Return takcluster task to trigger a buildbot builder.
+
+    This function creates a generic task with the minimum ammount of
```
gregarndt: 'amount'

mozci/sources/buildbot_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+    information required for the buildbot-bridge to consider it valid.
+    You can establish a list depencies to other tasks through the requires field.
```
gregarndt: dependencies

mozci/sources/buildbot_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+    :rtype: dict
+
+    """
+    task = {
+        'taskId': slugId(),
+        'reruns': 0,  # Do not retry the task if it fails to run successfuly
+        'task': {
+            'workerType': 'buildbot-bridge',
+            'provisionerId': 'buildbot-bridge',
+            'created': fromNow('0d'),
```
gregarndt, line 44: I'm surprised that there isn't a method that's more explicit in taskcluster-client for "now".  

If you don't like how this reads, you could just use the same datetime stuff we do in the mach command:
`datetime.datetime.utcnow().isoformat() + 'Z'`

mozci/sources/buildbot_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+                result = False
+                break
+
+    return result
+
+
+def generate_task_graph(repo_name, revision, builders_graph, **params):
+    """Return TaskCluster graph based on builders_graph.
+
+    :param revision: push revision
```
gregarndt: nit: didn't describe repo_name parameter here

mozci/sources/buildbot_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+            if repo_name not in dep_builder:
+                result = False
+                break
+
+    return result
+
+
+def generate_task_graph(repo_name, revision, builders_graph, **params):
+    """Return TaskCluster graph based on builders_graph.
+
```
gregarndt, line 108: nit: didn't describe repo_name parameter here

mozci/sources/buildbot_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+            break
+        for dep_builder in dep_builders:
+            if repo_name not in dep_builder:
+                result = False
+                break
+
+    return result
+
+
+def generate_task_graph(repo_name, revision, builders_graph, **params):
```
gregarndt: is params used for anything?

mozci/scripts/misc/buildbot_to_taskcluster.py
```diff
@@ -0,0 +1,59 @@
...
+
+    options = parser.parse_args()
+
+    if options.debug:
+        setup_logging(logging.DEBUG)
+    else:
+        setup_logging()
+
+    sch = BBBSchedulingClient()
+    sch.schedule_task_graph(
```
gregarndt: Am I understanding things correctly that this doesn't actually schedule the graph, but prints it out?  Also, If a graph couldn't be created, I think the docs say that "buildbot_bridge.generate_task_graph" could return None.  Should the user be given any indication that nothing was done?

mozci/sources/buildbot_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+
+    return task
+
+
+def _query_metadata(repo_name, revision):
+    repo_url = query_repo_url(repo_name)
+    push_info = query_revision_info(repo_url, revision)
+
+    return {
+        'owner': push_info['user'],
```
gregarndt, line 73: Is user always an email address?  The schema for `owner` states it should be an email [1].

I think there are times where user might just be a username of sorts (e.g. "user": "ffxbld").  This might not be an issue depending on how this is triggered.

[1] https://github.com/taskcluster/taskcluster-queue/blob/master/schemas/create-task-request.yml#L144

mozci/sources/buildbot_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+    metadata = _query_metadata(repo_name, revision)
+    _validate_builders_graph(repo_name, builders_graph)
+
+    # This is the initial task graph which we're defining
+    task_graph = {
+        'scopes': [
+            'queue:define-task:buildbot-bridge/buildbot-bridge',
+        ],
+        'tasks': [],
+        'metadata': dict(metadata.items() + {'name': 'task graph local'}.items())
```
gregarndt, line 134: is `task graph local` unique enough to know which task is what when viewing the graph in the inspector?  Wouldn't all tasks have this name?

mozci/scheduling.py
```diff
@@ -18,11 +23,99 @@ class BaseSchedulingClient:
...
+    @abstractmethod
+    def cancel_all(self, repo_name, revision, *args, **kwargs):
+        pass
+
+# End of BaseSchedulingClient
+
+
+class BuildAPISchedulingClient(BaseSchedulingClient):
+
+    def schedule_arbitrary_task(self, repo_name, revision, uuid, *args, **kwargs):
```
bhearsum: It's a bit weird to me that there's a method that schedules a task inside a BuildAPI class. BuildAPI doesn't know anything about Tasks. Perhaps this method needs a new name, since it triggering a Build, not a Task?

mozci/scheduling.py
```diff
@@ -18,11 +23,99 @@ class BaseSchedulingClient:
...
+        pass
+
+    @abstractmethod
+    def cancel_all(self, repo_name, revision, *args, **kwargs):
+        pass
+
+# End of BaseSchedulingClient
+
+
+class BuildAPISchedulingClient(BaseSchedulingClient):
```
bhearsum: I'm not sure that "SchedulingClient" really captures what this does. AFAICT, this is a general BuildAPI client, since canceling isn't considering scheduling.

mozci/scheduling.py
```diff
@@ -18,11 +23,99 @@ class BaseSchedulingClient:
...
+
+    def cancel(self, uuid, *args, **kwargs):
+        pass
+
+    def cancel_all(self, repo_name, revision, *args, **kwargs):
+        pass
+# End of TaskClusterSchedulingClient
+
+
+class BBBSchedulingClient(TaskclusterSchedulingClient):
```
bhearsum: This one is really throwing me off. MozCI is not scheduling things directly through BBB, it's submitting Task Graphs that happen to contain buildbot bridge jobs. This being a subclass of TaskclusterSchedulingClient seems like it's going to cause you issues as soon as you need a graph with both bridge and non-bridge jobs in it.

mozci/sources/buildbot_bridge.py
armenzg, line 44: I will check.

mozci/sources/buildbot_bridge.py
armenzg, line 73: It generally is an email address.
I'm OK with the graph being rejected.
I can figure it out once we find the case.

mozci/sources/buildbot_bridge.py
armenzg: Removed.

mozci/sources/buildbot_bridge.py
armenzg, line 108: Fixed.

mozci/sources/buildbot_bridge.py
armenzg, line 134: The tasks have their buildername as their name (line 57).
Only the graph has that name.

https://github.com/armenzg/mozilla_ci_tools/pull/337/files#diff-3e62fd7d392e6dd33a14da0de87c4348R57

> ```
>        'metadata': dict(metadata.items() + {'name': buildername}.items()),
> ```

mozci/scripts/misc/buildbot_to_taskcluster.py
armenzg: This is a random script just to test the POC.
We won't be using this entry point.

We will handle this better once mozci switches from using BuildAPI to this.

mozci/sources/buildbot_bridge.py
```diff
@@ -0,0 +1,160 @@
...
+    # XXX: We should validate the dependencies
+    result = True
+    for builder, dep_builders in builders_graph.iteritems():
+        if repo_name not in builder:
+            result = False
+            break
+        for dep_builder in dep_builders:
+            if repo_name not in dep_builder:
+                result = False
+                break
```
adusca, line 101: Can you validate here that everything has the same repo_name? Maybe having result be a set to which you append repo_names and in the end you return len(result) == 1?

Is this function also responsible for validating that a revision exists on a branch? How does TC deals with bad revisions?

Is it possible for dep_builder to be itself a graph with its own dep_builders? If so, maybe the function should be recursive (is this what the XXX is about? I didn't really catch what it means).

adusca:
Does BBB allow us to create a task that depends on a job that is already running?

mozci/scheduling.py
armenzg: I was trying to default to the new paradigm of tasks.
Otherwise, the base class cannot have a name that is good for both BuildAPI and TC.

mozci/scheduling.py
armenzg: I see what you mean.
It is more like task or job management.
Or CI client or CI manager.

mozci/scheduling.py
armenzg: Very good point. I will see what I can think of, however, I will be moving the functionality to another PR.
Thanks for the idea!

armenzg:
BBB does not allow us to add a dependency to a job which is already running.

Darn it! We will have to deal with that ad-hoc unless we push hard on moving scheduling away from buildbot. I think it is a better option that having to build something for only some quarters.

What do you think?

mozci/sources/buildbot_bridge.py
armenzg, line 101: I will deal with this on my next patch.
Do you see anything else that could be a blocker before landing this?

adusca:
No blockers, everything looks good to me.

armenzg:
I'm carrying forward the matters that still need addressing to issue #339 
If by next week the BBB blocker gets fixed, I would not to be in the middle of a refactoring.

If there's anything else you would like to carry forward please chime in.
Thank you all for your feedback!

armenzg:
I will focus a couple of days on Fx UI tests until I come back to this.

Issue #248, opened 2015-06-05
Raise less general Exceptions


adusca:
We currently raise a lot of general 'Exceptions' [1]. Instead, we should raise more specific Exceptions (like in [2]). This would help with issue #191.

Steps to fix:
- create a new Exception class [3], it may be necessary to create more than one
- replace every 'raise Exception' with the appropriate 'raise {insert_new_exception_name_here}'

[1]
mozilla_ci_tools[master $%]$ grep -nR "raise Exception" *
mozci/platforms.py:127:        raise Exception("No build job found.")
mozci/mozci.py:204:        raise Exception("The status of the job is expected to have a "
mozci/mozci.py:267:        raise Exception("Repository name not found in buildername. "
mozci/sources/pushlog.py:101:        raise Exception('Unable to retrieve pushlog data. '
mozci/sources/buildapi.py:162:    raise Exception("Not implemented because of bug 1087336. Use "
mozci/sources/buildapi.py:224:            raise Exception("Unexpected status")
mozci/sources/buildapi.py:262:            raise Exception("That repository does not exist.")
mozci/scripts/trigger.py:110:        raise Exception(error_message)
mozci/scripts/generate_triggercli.py:116:        raise Exception('The script supports only for fx-team, mozilla-inbound, mozilla-aurora')
mozci/utils/transfer.py:89:                    raise Exception(
mozci/utils/transfer.py:165:            raise Exception("We received %s which is unexpected." % req.status_code)
mozilla_ci_tools[master]$ 

[2] https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/sources/buildjson.py#L166
[3] https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/sources/buildjson.py#L22

armenzg:
We should also consider turning "assert" statements into raise customized exceptions.
Good idea adusca!

nikkisquared:
I'm working on this here https://github.com/nikkisquared/mozilla_ci_tools

nikkisquared:
I've pushed my progress so far. I'd like any feedback for if I'm approaching this in the right way, or if not, what I can do differently.

armenzg:
Commented on the PR.

armenzg:
Fixed by @nikkisquared

PR #342, opened 2015-09-04
Issue #248: Making custom, specific exceptions

nikkisquared:
Note that this pull request is not ready for merging, but I'm opening it now to ask for feedback. I started working on issue #248 as suggested by adusca. So far I've spent most of my time trying to figure out the level of specificity I should be making. For some edits I feel like it's just right, but others feel too specific. I'd really appreciate any feedback on what I'm doing right/wrong, and what I can do to improve my changes. Thanks!

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -14,6 +14,7 @@ from __future__ import absolute_import
 
 import logging
 
+from mozci.errors import MozciError
 from mozci.platforms import determine_upstream_builder, is_downstream, \
     filter_buildernames, build_talos_buildernames_for_repo
 from mozci.sources import allthethings, buildapi, buildjson, pushlog
@@ -212,7 +213,11 @@ def _status_info(job_schedule_info):
 
 
 def _find_files(job_schedule_info):
-    """Find the files needed to trigger a job."""
+    """
+    Find the files needed to trigger a job.
+
+    Raises MozciError if the job status doesn't have a properties key.
+    """
     files = []
 
     job_status = _status_info(job_schedule_info)
@@ -223,8 +228,8 @@ def _find_files(job_schedule_info):
 
     if not properties:
         LOG.error(str(job_status))
-        raise Exception("The status of the job is expected to have a "
-                        "properties key, however, it is missing.")
+        raise MozciError("The status of the job is expected to have a "
+                         "properties key, however, it is missing.")
 
     LOG.debug("We want to find the files needed to trigger %s" %
               properties["buildername"])
@@ -251,7 +256,11 @@ def query_builders():
 
 
 def query_repo_name_from_buildername(buildername, clobber=False):
-    """Return the repository name from a given buildername."""
+    """
+    Return the repository name from a given buildername.
+
+    Raises MozciError if there is no repository name in buildername.
+    """
     repositories = buildapi.query_repositories(clobber)
     ret_val = None
     for repo_name in repositories:
@@ -266,8 +275,8 @@ def query_repo_name_from_buildername(buildername, clobber=False):
         query_repo_name_from_buildername(buildername, clobber=True)
 
     if ret_val is None:
-        raise Exception("Repository name not found in buildername. "
-                        "Please provide a correct buildername.")
+        raise MozciError("Repository name not found in buildername. "
+                         "Please provide a correct buildername.")
 
     return ret_val
 
--- a/mozci/platforms.py
+++ b/mozci/platforms.py
@@ -4,6 +4,7 @@ import collections
 import logging
 import re
 
+from errors import MozciError
 from sources.allthethings import fetch_allthethings_data, list_builders
 
 LOG = logging.getLogger('mozci')
@@ -101,6 +102,8 @@ def determine_upstream_builder(buildername):
     When buildername corresponds to a test job it determines the
     triggering build job through allthethings.json. When a buildername
     corresponds to a build job, it returns it unchanged.
+
+    Raises MozciError if no matching build job is found.
     """
     _process_data()
 
@@ -130,7 +133,7 @@ def determine_upstream_builder(buildername):
     # "Android armv7 API 11+ larch build"
     if buildername.lower() not in BUILDERNAME_TO_TRIGGER:
         LOG.error("We didn't find a build job matching %s" % buildername)
-        raise Exception("No build job found.")
+        raise MozciError("No build job matching %s found." % buildername)
 
     shortname = BUILDERNAME_TO_TRIGGER[buildername.lower()]
     for suffix in SUFFIXES:
--- a/mozci/scripts/generate_triggercli.py
+++ b/mozci/scripts/generate_triggercli.py
@@ -112,7 +112,8 @@ def check_repository(buildername):
     supported_repositories = ['fx-team', 'mozilla-inbound', 'mozilla-aurora']
     repo_name = query_repo_name_from_buildername(buildername)
     if repo_name not in supported_repositories:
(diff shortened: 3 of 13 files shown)


mozci/mozci.py
```diff
@@ -40,6 +40,14 @@
 QUERY_SOURCE = BuildApi()
 
 
+class JobException(Exception):
```
armenzg: For the record, I have no clear view of what right should look like.
Let's pick something that it more or less feels right :)

Can you change these two exceptions to be MozciException? It makes implicit reference to the module this code lives within (mozci.py) and it is not as specific to the actual issue raised.

mozci/platforms.py
```diff
@@ -9,6 +9,10 @@
 LOG = logging.getLogger('mozci')
 
 
+class BuildException(Exception):
```
armenzg: Could you import the MozciException from the other module?
platforms.py was kind of an extension of mozci.py

mozci/scripts/generate_triggercli.py
```diff
@@ -44,6 +44,14 @@
 LOG = setup_logging()
 
 
+class RepoNameException(Exception):
+    pass
+
+
+class BugCommentException(Exception):
```
armenzg: Let's ignore all scripts completely.
All files under mozci/scripts/ are not components that external consumers will be using.
The scripts/ directory lives under mozci instead of the root directory because the usage we have of console_scripts under setup.py which requires us to put it under mozci/
One day I hope we will get rid of that.

mozci/scripts/trigger.py
```diff
@@ -13,6 +13,10 @@
 from mozci.sources.pushlog import query_repo_tip
 
 
+class ValidationException(Exception):
```
armenzg: Let's ignore this script as well.

mozci/sources/buildapi.py
```diff
@@ -31,6 +31,10 @@ class BuildapiException(Exception):
     pass
 
 
+class RepoNameException(Exception):
```
armenzg: Let's call it BuildApiException.

mozci/sources/pushlog.py
```diff
@@ -27,6 +27,10 @@
 VALID_CACHE = {}
 
 
+class PushlogException(Exception):
```
armenzg: Good :+1:

mozci/utils/transfer.py
```diff
@@ -26,6 +26,10 @@
 MEMORY_SAVING_MODE = False
 
 
+class StatusException(Exception):
```
armenzg: Let's import the MozciException.

Hrmm I never realized how complicated naming exceptions is.

mozci/sources/buildapi.py
armenzg: I realized we already have the BuildapiException inside of this module. Can you please use that?

armenzg:
Hi @nikkisquared, thank you for this patch!

Could you also try to add information about the exceptions in the docstrings where it applies?
In the pep 257 docs it is suggested [1]
My apologies about the modules not being properly documented according to that standard but I would like to add info about the exceptions so consumers can be aware of it (even if the rest of the docstring is not complete).

[1] https://www.python.org/dev/peps/pep-0257/
The docstring for a function or method should summarize its behavior and document its arguments, return value(s), side effects, exceptions raised, and restrictions on when it can be called (all if applicable)

mozci/platforms.py
vaibhavmagarwal: This will cause a circular import since platforms.py is already imported in mozci.py
We can define the MozciException here once again or let this remain as a normal base exception.

vaibhavmagarwal: Or we can introduce another file "errors.py" and define MozciException there and import it in all the places.

armenzg:
Having mozci/exceptions.py would be useful.

nikkisquared:
I've just pushed my second draft, following all of your suggestions for the names and docstring usage. Thanks a lot for the advice @armenzg, you answered pretty much every question I had in mind! The next thing for me to do would be writing documentation for the exceptions I haven't done anything to, and possibly putting all the custom exceptions in mozci/exceptions.py.

There's two exceptions I haven't done anything with, which are the "_valid_builder" function in mozci/sources/buildapi.py that instantly raises an error when called, and the exception in "_load_json_file" mozci/utils/transfer.py that only comes up if a Windows user doesn't have gzip installed. What should I do about either of those, if anything?

armenzg:
For [1], it perhaps make sense to simply say "pass" and turn the exceptions message to be a simple comment. We're simply not to use that method. We could even remove it completely.
The exception perhaps was unneccesarily dramatic :)

For [2], I doubt we will have tools being built using Windows that would require us to promise them a known exception.

For reference:
[1] https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/sources/buildapi.py#L156
[2] https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/utils/transfer.py#L104

armenzg:
The PR looks good to me.

WRT to "@ adusca armenzg vaibhav1994 I think I should put "Error" instead of "Exception" in each of my custom exceptions, following what Python and some existing code does"

Do you have a reference to using Error instead?
I'm happy either way but I would like to know what you're refering to.

armenzg:
@chmanchester since you requested Mozci to have more formalized exceptions (rather than generic ones), would you mind having a look to see if this meets your expectations? Thanks!

nikkisquared:
@armenzg I should've linked what I was looking at in my last comment, sorry. The standard python exceptions all use Error in the name https://docs.python.org/2/c-api/exceptions.html#standard-exceptions and I noticed that sources/buildapi.py imports AuthenticationError from another mozci module. So it seems like changing Exception to Error would be good practice. Also @vaibhavmagarwal (whose comments here I can't seem to find again) originally suggested the Exceptions.py module be called Errors.py, and I hadn't thought about why that might be until now.

nikkisquared:
I've renamed all the exceptions to use Error now, and removed the the "_valid_builder" function.

I made sure my work was done with `grep "(Exception) * -r` and found that there are three custom-named Exceptions in the source code [1] [2] [3] that I could put in errors.py and import from there. I could also re-name the latter two to be "Error" instead of "Exception. Thoughts @armenzg?

[1] https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/query_jobs.py#L17
[2] https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/sources/buildjson.py#L22
[3] https://github.com/armenzg/mozilla_ci_tools/blob/master/mozci/utils/authentication.py#L20

armenzg:
Thanks for clarifying about the error naming. It makes sense.

Please go ahead and move those exceptions into the errors module and rename them properly.
We can merge after that.

nikkibee: would you be interested on contributing in other issues as well?
Also, would you mind making a mozillians.org profile once you have a chance? It's easier to help Mozilla reach out to you. I can vouch for you.

armenzg:
Hi Nikkibee,
Could you also try to squash your commits?
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

I would like to merge one commit instead of multiple ones.
I have not done this myself but I have been aware that it is a good practice which I would like to get started.
Let me know if you have trouble.

nikkisquared:
@armenzg I'll move the other three exceptions into errors.py and then see about squashing the commits. Then I'll make a mozillians account. Thanks again for all the feedback!

chmanchester:
These exceptions look good to me, thanks for the patch!

nikkisquared:
Your welcome @chmanchester! I've enjoyed working on it.

PR #349, opened 2015-09-29
Issue 339 - Schedule task graphs directly + dry_run + singole builder

armenzg:
## ci_manager.py
- Rename TaskclusterSchedulingClient to TaskClusterBuildbotManager
- Move graph scheduling logic to the tc module
- Support passing along _args and *_kwargs (for dry_run support)
- TaskClusterBuildbotManager can now schedule a single builder.
## buildbot_to_taskcluster.py
- Shorter help text on one of the options
- Pass the dry_run value
- We can now schedule single builders
## buildbot_bridge.py
- generate_graph_from_builders() can now receive _args and *_kwargs
- Added scheduler:create-task-graph scope to the task_graph template
- Added schedule_graph() method which calls the tc client's scheduler
## authentication.py
- Use path_to_file to have the credentials on the same spot as every other auxiliary file
- Store password as 'ldap' instead of 'mozci' in the keyring

### DIFF
--- a/mozci/sources/buildbot_bridge.py
+++ b/mozci/sources/buildbot_bridge.py
@@ -103,6 +103,23 @@ def _validate_builders_graph(repo_name, builders_graph):
     return result
 
 
+def generate_graph_from_builder(repo_name, revision, buildername, *args, **kwargs):
+    """Return TaskCluster graph based on buildername.
+
+    :param repo_name The name of a repository e.g. mozilla-inbound, alder et al.
+    :type repo_name: str
+    :param revision: push revision
+    :type revision: str
+    :param buildername: Buildbot buildername
+    :type revision: str
+
+    :returns: return None or a valid taskcluster task graph.
+    :rtype: dict
+
+    """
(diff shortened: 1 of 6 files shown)


mozci/sources/buildbot_bridge.py
```diff
@@ -103,6 +103,23 @@ def _validate_builders_graph(repo_name, builders_graph):
...
+    :param revision: push revision
+    :type revision: str
+    :param buildername: Buildbot buildername
+    :type revision: str
+
+    :returns: return None or a valid taskcluster task graph.
+    :rtype: dict
+
+    """
+    return generate_task_graph(repo_name, revision, {buildername: []})
```
adusca, line 120: I'm not sure I understand what is going on. What happens if we pass a downstream job whose upstream job doesn't exist/is not finished? The graph is created? BBB can process it? Do we get an error?

armenzg, line 120: It will probably won't work. I only made some progress in that area for experimenting purposes.
I would prefer to get it in for now and work on it more if someone has a use for it.

PR #356, opened 2015-10-06
Bug 1212004 - Handle when buildapi fails to submit a request

armenzg:
r?

We could have also added retry logic for now this is good enough (since at least we can recognize the status error).

https://bugzilla.mozilla.org/show_bug.cgi?id=1212004

### DIFF
--- a/mozci/sources/buildapi.py
+++ b/mozci/sources/buildapi.py
@@ -44,18 +44,23 @@ def trigger_arbitrary_job(repo_name, builder, revision, files=[], dry_run=False,
         LOG.info("         with this payload: %s" % str(payload))
         return None
 
-    # NOTE: A good response returns json with request_id as one of the keys
-    req = requests.post(
-        url,
-        headers={'Accept': 'application/json'},
-        data=payload,
-        auth=get_credentials()
-    )
-    if req.status_code == 401:
-        remove_credentials()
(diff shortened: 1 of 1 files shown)


vaibhavmagarwal:
From the traceback, I think the ValueError is occurring only because of this line: `content = req.json()`
We are not using the `content` variable in the function, so can we just remove it? I think that will fix the ValueError, as the response may not always be a correctly formatted json.

armenzg:
Hi @vaibhavmagarwal you're right that the exceptions comes from content = req.json(), however, I believe I moved the whole block since I was considering adding more exceptions from the post call (which I didn't).

Let me fix this.

contents is used a little below:
https://github.com/armenzg/mozilla_ci_tools/pull/356/files#diff-e9eac7c08694fea7da57a886c52ff7d6R64

armenzg:
Is this better? r?

vaibhavmagarwal:
r+
Just another query, should we retry posting it once instead of directly exiting? I believe that can be a bit better, but upto you.

armenzg:
I've considered it but I did not want to go there since I really dislike the retry logic of requests.

I've considered adding the redo python package as a dependency but I did not want to increase the scope.

I will take this as-is as I want to bump the mozci version for another fix for pulse_actions.

PR #358, opened 2015-10-06
Issue 357 - Coalesced builds should be the same as non-existant builds

armenzg:
In the for loop where we try to evaulate if there is a build for a given
push we don't detect properly that a coalesced build should not count
as an existant build.

This causes us to run test jobs on builds for the push that actually
run the build.

This causes the tests to be run against a wrong build.

This code change ignores coalesced builds.

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -134,14 +134,21 @@ def _determine_trigger_objective(revision, buildername, trigger_build_if_missing
         try:
             status = query_api.get_job_status(job)
         except buildjson.BuildjsonException:
-            LOG.debug("We have hit bug 1159279 and have to work around it. We will pretend that "
-                      "we could not reach the files for it.")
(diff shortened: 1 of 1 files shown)


mozci/mozci.py
```diff
@@ -134,16 +134,21 @@ def _determine_trigger_objective(revision, buildername, trigger_build_if_missing
...
         # Sometimes running jobs have status unknown in buildapi
-        if status == RUNNING or status == PENDING or status == UNKNOWN:
+        if status in (RUNNING, PENDING, UNKNOWN):
             LOG.debug("We found a running/pending build job. We don't search anymore.")
             running_job = job
             # We cannot call _find_files for a running job
             continue
 
+        # Having a coalesced build is the same as not having a build available
+        if status == COALESCED:
```
chmanchester, line 149: It might be helpful to log a DEBUG message here.

armenzg, line 149: Added

PR #360, opened 2015-10-06
Issue 359 - Prompt user for storing password + fix storing key

armenzg:
r?

We can store a user's password in the keyring in an encrypted manner.
Unfortunately, we were not storing the password after prompting the
user.

We also fix in here a small regression where we started storing
the password with 'ldap' as the key instead of 'mozci'.

### DIFF
--- a/mozci/utils/authentication.py
+++ b/mozci/utils/authentication.py
@@ -7,16 +7,68 @@
 """Module for http authentication operations."""
 
 import getpass
-import keyring
-import logging
 import os
 
-from mozci.utils.transfer import path_to_file
+import keyring
+import requests
 
+from mozci.utils.transfer import path_to_file
+from mozci.utils.log_util import setup_logging
+
+AUTH = None
 CREDENTIALS_PATH = path_to_file("credentials.cfg")
 DIRNAME = os.path.dirname(CREDENTIALS_PATH)
-LOG = logging.getLogger('mozci')
-AUTH = None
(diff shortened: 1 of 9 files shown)


mozci/utils/authentication.py
```diff
@@ -52,25 +63,21 @@ def get_credentials():
     if content is not None:
         LOG.debug("Loading LDAP user from %s" % CREDENTIALS_PATH)
         https_username = content[0].strip()
-        https_password = keyring.get_password("ldap", https_username)
+        https_password = keyring.get_password(KEYRING_KEY, https_username)
         if https_password is None or https_password == "":
             https_password = getpass.getpass(
                 "Input LDAP password for user %s: " % https_username)
+            _prompt_password_storing(https_username, https_password)
```
vaibhavmagarwal: We did not prompt for storing password above because there was a specific request by developers who did not want to store passwords on certain machines. It will be annoying to them if we prompt each time for storing the password, even when they have said "no" initially.

armenzg:
Hi @vaibhavmagarwal I've taken what you mentioned into consideration.

I was struggling because I had already had said 'no' a while ago but I could not tell from the output.
We don't ask the user anymore but we mention in the output what to do if they have changed their mind.

This PR fixes the wrong keying key + it fixes the authentication's logging.

I decided to not switch everywhere to use "setup_logging" even though we could have.

vaibhavmagarwal:
Looks good. r+

PR #378, opened 2015-11-16
Issue 376: Update buildbot_graph_builder so it can take any list of jobs

adusca:
buildbot_graph_builder will take a list of jobs that we want to trigger and
a revision, and create a graph with the necessary build jobs.

r? @armenzg

### DIFF
--- a/mozci/sources/buildbot_bridge.py
+++ b/mozci/sources/buildbot_bridge.py
@@ -5,8 +5,8 @@ which will use the buildbot-bridge system to trigger them on buildbot.
 from __future__ import absolute_import
 
 from mozci.errors import MozciError
-from mozci.mozci import valid_builder
-from mozci.platforms import get_buildername_metadata
+from mozci.mozci import determine_trigger_objective, valid_builder
+from mozci.platforms import is_downstream, get_buildername_metadata
 from mozci.sources.buildapi import query_repo_url
 from mozci.sources.pushlog import query_revision_info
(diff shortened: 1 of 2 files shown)


mozci/sources/buildbot_bridge.py
```diff
@@ -118,13 +118,42 @@ def buildbot_graph_builder(builders):
...
     """
     graph = {}
+
+    # We need to determine what upstream jobs need to be triggered besides the
+    # builders already on our list
     for b in builders:
-        graph[b] = None
+        # for test jobs, determine_trigger_objective[0] is the build
+        # job when no build job exists the test job when the build job
+        # is completed and None when the build job is running
```
armenzg: Could you please structure this paragraph as a bullet point?
It is a bit hard to read and understand.

mozci/sources/buildbot_bridge.py
```diff
@@ -118,13 +118,42 @@ def buildbot_graph_builder(builders):
...
+        if is_downstream(b):
+            objective = determine_trigger_objective(revision, b)[0]
+
+            if objective == b:
```
armenzg, line 140: Please add a note in hear to understand what we're doing.
Same below.

armenzg:
Looks good to me. Please merge when you address the nits mentioned.

adusca:
Thanks for the review! I hope the new comments make more sense.

PR #380, opened 2015-11-18
Add --existing-only support to triggerbyfilters

xennygrimmato:
If the user passes --existing-only option while running triggerbyfilters.py, jobs with existing builds get triggered.

### DIFF
--- a/mozci/scripts/triggerbyfilters.py
+++ b/mozci/scripts/triggerbyfilters.py
@@ -1,6 +1,5 @@
 """
 This script retriggers N times every job that matches --includes and doesn't match --exclude .
-
 usage:
 python th_filters.py repo rev --includes "args to include" --exclude  "args to exclude --times N"
 """
@@ -71,6 +70,11 @@ def parse_args(argv=None):
                         default="buildapi",
                         help="Query info from buildapi or treeherder.")
 
+    parser.add_argument("--existing-only",
+                        action="store_false",
+                        dest="existing_only",
+                        help="Only trigger test jobs if the build jobs already exists.")
+
     options = parser.parse_args(argv)
     return options
 
@@ -108,8 +112,13 @@ def main():
         LOG.info("0 jobs match these filters, please try again.")
         return
 
-    cont = raw_input("%i jobs will be triggered, do you wish to continue? y/n/d (d=show details) "
-                     % len(buildernames))
+    if options.existing_only:
+        cont = raw_input("The ones which have existing builds out of %i jobs will be triggered,\
+                         do you wish to continue? y/n/d (d=show details) " % len(buildernames))
+    else:
+        cont = raw_input("%i jobs will be triggered, do you wish to continue? \
+                          y/n/d (d=show details) " % len(buildernames))
+
     if cont.lower() == 'd':
         LOG.info("The following jobs will be triggered: \n %s" % '\n'.join(buildernames))
         cont = raw_input("Do you wish to continue? y/n ")
@@ -119,21 +128,24 @@ def main():
 
     # Setting the QUERY_SOURCE global variable in mozci.py
     set_query_source(options.query_source)
-
     for buildername in buildernames:
+        repo_name = query_repo_name_from_buildername(buildername)
+
+        trigger_build_if_missing = True
+        if options.existing_only or repo_name == 'try':
+            trigger_build_if_missing = False
+
         trigger_range(
             buildername=buildername,
             revisions=[revision],
             times=options.times,
             dry_run=options.dry_run,
+            trigger_build_if_missing=trigger_build_if_missing,
         )
 
-        LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
-                 urllib.urlencode({'repo': query_repo_name_from_buildername(buildername),
-                                   'fromchange': revision,
-                                   'tochange': revision,
-                                   'filter-searchStr': buildername}))
-
+    LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
+             urllib.urlencode({'repo': options.repo,
+                               'revision': revision}))
 
 if __name__ == '__main__':
     main()


mozci/scripts/triggerbyfilters.py
```diff
@@ -21,7 +21,6 @@
 from mozci.sources.buildapi import query_repo_url
 from mozci.sources.pushlog import query_repo_tip
 
-
```
vaibhavmagarwal, line 27: Could you please put the line back?

mozci/scripts/triggerbyfilters.py
```diff
@@ -67,13 +66,19 @@ def parse_args(argv=None):
...
+                        dest="existing_only",
+                        help="Only trigger test jobs if the build jobs already exists.")
+    
     options = parser.parse_args(argv)
     return options
 
 
 def main():
     options = parse_args()
+
```
vaibhavmagarwal: Could you please remove the above extra line.

mozci/scripts/triggerbyfilters.py
```diff
@@ -114,17 +119,26 @@ def main():
 
     # Setting the QUERY_SOURCE global variable in mozci.py
     set_query_source(options.query_source)
-
+
```
vaibhavmagarwal: Could you please remove the line and extra whitespace?
Tip: I have added settings in my editor to automatically remove trailing white space on saving.

mozci/scripts/triggerbyfilters.py
```diff
@@ -114,17 +119,26 @@ def main():
...
-
+    
     for buildername in buildernames:
+        repo_name = query_repo_name_from_buildername(buildername)
+        
+        trigger_build_if_missing = True
+        if options.existing_only or repo_name=='try':
+            trigger_build_if_missing = False
+        else:
+            trigger_build_if_missing = True
```
vaibhavmagarwal: The "else" part of the statement is not required as "trigger_build_if_missing" will remain True when the if statement is not executed.

mozci/scripts/triggerbyfilters.py
```diff
@@ -114,17 +119,26 @@ def main():
...
+                
         trigger_range(
             buildername=buildername,
             revisions=[options.rev],
             times=options.times,
             dry_run=options.dry_run,
+            trigger_build_if_missing=trigger_build_if_missing,
         )
-
+
```
vaibhavmagarwal: Remove the extra white space.

vaibhavmagarwal:
@xennygrimmato thanks for working on this! Just did a drive-by and added some comments :)
Will leave the actual review to @adusca, as she knows what should be the default feature.

xennygrimmato:
@vaibhavmagarwal I have removed the extra whitespaces and newlines in 5ce165c . Thanks for pointing the un-required "else" part :)

adusca:
@vaibhavmagarwal already covered everything :)

This looks right to me. Unfortunately, there are merge conflicts :/. @xennygrimmato Can you fix the conflicts so I can test locally before merging? Also, if you could add a note in the documentation about the new flag: https://github.com/mozilla/mozilla_ci_tools/blob/master/docs/scripts.rst#triggerbyfilterspy

Thanks for the PR!

xennygrimmato:
@adusca I have fixed merge conflicts. Can you please test it locally? I will add documentation about the flag tonight.

armenzg:
Looking at this now.

armenzg:
From the output, it is very hard to determine what happened.
Could you please look into modifying the mozci method we call to improve the output?

If you run the command below with --debug you will see more of what is going on.
11/20/2015 12:09:39 mozci DEBUG:     Nothing needs to be triggered

I'm using this push without dry-run since it has failed Mac builds:
https://treeherder.mozilla.org/#/jobs?repo=alder&revision=468fe389a811

(ci_tools)armenzg@armenzg-thinkpad:~/repos/ci_tools$ mozci-triggerbyfilters -i "Yosemite,talos,g1" --existing-only alder 468fe389a811 
11/20/2015 12:05:29 root INFO:   Setting INFO level
1 jobs will be triggered, do you wish to continue? y/n/d (d=show details) y
11/20/2015 12:05:34 mozci INFO:  We want to have 1 job(s) of Rev5 MacOSX Yosemite 10.10 alder talos g1 on revisions [u'468fe389a8117be31db0aed7aad0d1cfa794372e']
11/20/2015 12:05:34 mozci INFO:  
11/20/2015 12:05:34 mozci INFO:  === 468fe389a8117be31db0aed7aad0d1cfa794372e ===
11/20/2015 12:05:35 mozci INFO:  We want to have 1 job(s) of Rev5 MacOSX Yosemite 10.10 alder talos g1 on revision 468fe389a8117be31db0aed7aad0d1cfa794372e
11/20/2015 12:05:36 mozci INFO:  We have found 0 potential job(s) matching 'Rev5 MacOSX Yosemite 10.10 alder talos g1' on 468fe389a8117be31db0aed7aad0d1cfa794372e. We need to trigger more.
11/20/2015 12:05:36 mozci INFO:  ===> We want to trigger 'Rev5 MacOSX Yosemite 10.10 alder talos g1' on revision '468fe389a8117be31db0aed7aad0d1cfa794372e' a total of 1 time(s).
11/20/2015 12:05:36 mozci INFO:  
builds-2015-09-15.js: [>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>] Elapsed Time: 0:00:09   3.59 MB/s 31.0MB
11/20/2015 12:05:52 mozci INFO:  We found a job that finished but it did not produced files. status: 6
11/20/2015 12:05:52 mozci INFO:  The build job OS X 10.7 alder build failed on revision 468fe389a8117be31db0aed7aad0d1cfa794372e without generating the necessary files. We will not trigger anything.
11/20/2015 12:05:52 root INFO:   https://treeherder.mozilla.org/#/jobs?repo=alder&fromchange=468fe389a8117be31db0aed7aad0d1cfa794372e&tochange=468fe389a8117be31db0aed7aad0d1cfa794372e&filter-searchStr=Rev5+MacOSX+Yosemite+10.10+alder+talos+g1

adusca:
I tested it locally, and thinks are working as expected :-) . I ran:

```
python mozci/scripts/triggerbyfilters.py try 00dcfb370692 -i crash --dry-run
```

On this revision: https://treeherder.mozilla.org/#/jobs?repo=try&revision=00dcfb370692

The original list of jobs that might be triggered was not changed, but it skipped every downstream job for which no build job existed. In the end it would have triggered only the 4 downstream jobs that we wanted (on master we trigger the 4 jobs we want and additional **28** build jobs!!!).

Output: https://pastebin.mozilla.org/8852795 (triggerbyfilters output is quite verbose, it is going to be way better after we transition to scheduling with BBB)
On master: https://pastebin.mozilla.org/8852796

adusca:
On here: https://github.com/mozilla/mozilla_ci_tools/pull/380/files#diff-fd91cda9284f16c1f7364ee281591b09R115

We said that the jobs will be triggered. Can you add an if/else block (depending on repo_name == try or options.existing_only) that says that only the jobs for which the build job exists are going to be triggered?

xennygrimmato:
Yes, I will do that.
I must also change the following lines:
https://github.com/xennygrimmato/mozilla_ci_tools/blob/95dbe43a3330949f560696b9667a1dbfa35555e4/mozci/scripts/triggerbyfilters.py#L140-L144

To show only triggered jobs in log, is that correct?
As per current behaviour, jobs that are not triggered will also be shown in the log. Whereas, we must show only the triggered ones. Thoughts?

adusca:
Good catch! I think in this case the best choice is to just add a link to the whole repo/revision view [1] outside the loop, instead of one link per buildername.

[1]

```
LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
                 urllib.urlencode({'repo': repo_name,
                                            'revision': revision}))
```

xennygrimmato:
We are currently saying that:
`
cont = raw_input("%i jobs will be triggered, do you wish to continue? y/n/d (d=show details) "
                     % len(buildernames))`
Since, only existing jobs will be triggered, `len(buildernames)` is the maximum number of jobs that may be triggered. Some times, fewer jobs will be triggered. Is that correct?
In that case, we must probably loop through all buildernames once and find the exact number of jobs that will be triggered.

vaibhavmagarwal:
@xennygrimmato yes, it is correct that there can be fewer jobs when existing only flag is used. It is not possible to know the exact number of jobs that will be triggered in this case without actually querying the APIs. An easier way out that @adusca is saying is, we should just prompt the users so as to not arise any confusion and hence use if/else statement. Something like:

```
if options.existing_only:
    cont = raw_input("The ones which have existing build out of %i jobs will be triggered, do you wish to continue? y/n/d (d=show details) "
                      % len(buildernames))
```

mozci/scripts/triggerbyfilters.py
```diff
@@ -119,21 +123,25 @@ def main():
 
     # Setting the QUERY_SOURCE global variable in mozci.py
     set_query_source(options.query_source)
-
     for buildername in buildernames:
+        repo_name = query_repo_name_from_buildername(buildername)
+
+        trigger_build_if_missing = True
+        if options.existing_only or repo_name == 'try':
```
vaibhavmagarwal, line 135: Just noticed this, the jobs on try with no existing builds will never be triggered since we are always making `trigger_build_if_missing` to `False` when repository is try. @adusca is that what we want?

xennygrimmato:
Since we aren't showing log of each buildername, we can remove the "d=show details" option right?

```
if cont.lower() == 'd':
        LOG.info("The following jobs will be triggered: \n %s" % '\n'.join(buildernames))
        cont = raw_input("Do you wish to continue? y/n ")
```

Because now we'll only be displaying only the repo name and revision, outside the loop:

```
LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
               urllib.urlencode({'repo': options.repo,
                               'revision': revision}))
```

Can you confirm this?

vaibhavmagarwal:
I think we will still need the 'd' option to show the names of all builders, since developers may want to see the potential builders when --existing-only flag is not used.

xennygrimmato:
Oh, that's right :) Thanks for pointing it out.

xennygrimmato:
I have pushed the changes. Can you please review it once?

mozci/scripts/triggerbyfilters.py
```diff
@@ -117,23 +126,26 @@ def main():
     if cont.lower() != 'y':
         exit(1)
 
+    LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
```
armenzg: Could you please move this block after the for/loop? That way the user will be able to see it at the end of the output.

mozci/scripts/triggerbyfilters.py
armenzg, line 135: I think this was a preventive measure since we had devs complaining that they added too many unnecessary builds.

I wonder if we should change the default to be "existing only" and instead have a flag --add-also-missing-builds.

I say this since on 'try' we would not be able to add missing builds unless the dev figures out that they need to specify the build for those tests with -i

What do you think?

vaibhavmagarwal, line 135: I think changing the default here with --add-also-missing-builds will alter the behavior  and make it the exact opposite of the current behavior. I am not a user of the scripts, but the developers may not like the change. What if we do not make it default on try, and the developers who do not want to trigger unnecessary builds use the flag --existing-only?

I think we can only speculate what our users want, ultimately the decision what should be the default lies with the users.

adusca, line 135: Triggering extra builds by default on try is not a good default behaviour.

We received complaints about this for alltalos, and we changed the behaviour there, now that we deprecated alltalos we are asking our users to be dance around filters to avoid triggering unnecessary builds (e.g. --exlclude win,linux,ubuntu,rev to only trigger android jobs), or remember another a flag.

armenzg, line 135: Let's only add more jobs if we're explicit about it.

On 24 November 2015 at 16:13, Alice notifications@github.com wrote:

> In mozci/scripts/triggerbyfilters.py
> https://github.com/mozilla/mozilla_ci_tools/pull/380#discussion_r45795522
> :
> 
> > ```
> >  for buildername in buildernames:
> > ```
> > -        repo_name = query_repo_name_from_buildername(buildername)
> >   +
> > -        trigger_build_if_missing = True
> > -        if options.existing_only or repo_name == 'try':
> 
> Triggering extra builds by default on try is not a good default behaviour.
> 
> We received complaints about this for alltalos, and we changed the
> behaviour there, now that we deprecated alltalos we are asking our users to
> be dance around filters to avoid triggering unnecessary builds (e.g.
> --exlclude win,linux,ubuntu,rev to only trigger android jobs), or remember
> another a flag.
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/mozilla/mozilla_ci_tools/pull/380/files#r45795522.

## 

Zambrano Gasparnian, Armen
Automation & Tools Engineer
http://armenzg.blogspot.ca

armenzg:
Hi @xennygrimmato , I'm very sorry I have not noticed you pushed more code.

Unless I see a comment after a code push I won't get notified.
I will have a look at this shortly.

xennygrimmato:
Thanks. Do notify me of the changes here.

armenzg:
This is good. Thank you!

PR #388, opened 2015-11-23
Make fill-in-a-revision work with TaskCluster #387

F3real:


mozci/sources/buildbot_bridge.py
```diff
@@ -23,9 +23,14 @@
 """
 from __future__ import absolute_import
 
+from mozci.ci_manager import TaskClusterBuildbotManager
```
adusca: So, turns out this is a circular import. It is possible to fix this by leaving the import inside the function (we would have to add an exception to flake8). I think this is the best option. @armenzg what do you think?

adusca:
This patch does exactly what I suggested. Unfortunately, there were some problems with my suggestion that I only realized running the code.

First, buildbot_task_builder is too slow. We could try to make it faster by adding a smart cache in determine_trigger_objective (so it only searches for a build job once). Also, since it calls detemine_trigger_objective and we do a lot of logging in determine_trigger_objective, it is veeeery verbose. We could move a lot of the logging on determine_trigger_objective to trigger_range. 

Another idea would be to not use buildbot_task_builder and instead write a function that builds the task graph while it runs through the existing jobs. This will be harder to write, but it has the potential of being very efficient.

Another problem that I saw, but that I have no idea of what is going on is that we are finding a lot of buildernames that are not of either coalesced jobs or existing jobs: around 700 per push. A lot of Windows XP pgo jobs on inbound, for example. @armenzg do you know what is going on?

mozci/sources/buildbot_bridge.py
armenzg: Interesting solution!
We need to query before we can determine what to schedule.

Try to put the logic inside of the ci_manager instead of buildbot_bridge.

This is how the logic might work:
- From query_source determine what could have been scheduled (you might need to add the functionality)
  - Something like builders_for_repo = list_builders(repo_name=repo_name)
- From query_source determine what jobs have run
- Substract to determine which jobs have not run
- Call buildbot_graph_builder() with the jobs that are missing (see block below)
- Schedule the jobs

uuid represents builders or jobs (list of unique identifiers).

```
    task_graph = buildbot_bridge.generate_graph_from_builders(
        repo_name=repo_name,
        revision=revision,
        buildernames=[uuid],
        *args, **kwargs
    )
    return super(TaskClusterBuildbotManager, self).schedule_graph(
        task_graph=task_graph, *args, **kwargs)
```

The problem you're facing here is that many of the sophisticated pieces of code under mozci/mozci.py are not currently available to you. The idea is that little by little we will move some of these pieces of codes from mozci/mozci.py to mozci/ci_manager.py. We will have a function per scheduling system. Eventually I want to remove the mozci.py file.

mozci/scripts/trigger.py
```diff
@@ -12,11 +12,11 @@
     query_repo_name_from_buildername,
     query_repo_url_from_buildername,
     set_query_source,
-    trigger_missing_jobs_for_revision,
     trigger_range,
 )
 from mozci.query_jobs import BuildApi, COALESCED
 from mozci.repositories import query_repo_url
+from mozci.sources.buildbot_bridge import trigger_missing_jobs_for_revision
```
armenzg: It would be great if we exposed the trigger_missing_jobs_for_revision functionality through the BBB class in query sources rather than importing buildbot_bridge directly.

I would prefer if we have an easy way in this script to use either BuildAPI or TaskCluster (--scheduling-manager='buildapi').
My apologies, I should have been clear in the requirements.

mozci/query_jobs.py
```diff
@@ -141,6 +141,27 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
         buildernames = right_status_buildernames - wrong_status_buildernames
         return sorted([request_id_by_buildername[b] for b in buildernames])
 
+    def remove_buildernames_without_status(self, repo_name, revision, status, buildernames):
```
armenzg: Could you please reword this to "select_jobs_with_status()"?
It's a bit clearer. Good job on making it generic!

armenzg:
Hi @F3real, good job here but it seems that this will require wider range of changes.
Please let me know if my big comment did not make sense.

regards,
Armen

mozci/mozci.py
```diff
@@ -501,27 +501,6 @@ def trigger(builder, revision, files=[], dry_run=False, extra_properties=None):
                                  extra_properties)
 
 
-def trigger_missing_jobs_for_revision(repo_name, revision, dry_run=False):
```
armenzg: I would like to keep the function available both for buildapi and buildbot_bridge.

mozci/query_jobs.py
```diff
@@ -141,6 +141,27 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
         buildernames = right_status_buildernames - wrong_status_buildernames
         return sorted([request_id_by_buildername[b] for b in buildernames])
 
+    def select_jobs_with_status(self, repo_name, revision, status, buildernames):
+        """
+        Removes all buildernames from 'buildernames' that have don't have jobs with 'status'.
```
armenzg: "with the specified 'status'" (reads a bit better)

mozci/query_jobs.py
```diff
@@ -141,6 +141,27 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
         buildernames = right_status_buildernames - wrong_status_buildernames
         return sorted([request_id_by_buildername[b] for b in buildernames])
 
+    def select_jobs_with_status(self, repo_name, revision, status, buildernames):
```
armenzg: Let's make "buildername" be named "considered_list_of_builders" and set to optional.

Let's also add a new function called "determine_missing_jobs" which calls this function.

mozci/query_jobs.py
```diff
@@ -141,6 +141,27 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
...
+        """
+        Removes all buildernames from 'buildernames' that have don't have jobs with 'status'.
+
+        Returns a list with the buildernames of the jobs whose only status isn't 'status'.
```
armenzg: Let's document the variables in here.
Let's mention that "considered_list_of_builders" can be used as a way to not consider _all_ possible builders for a repository since many builders are not scheduled (e.g. pgo build jobs on inbound).

mozci/query_jobs.py
```diff
@@ -141,6 +141,27 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
...
+        """
+        all_jobs = self._get_all_jobs(repo_name, revision)
+        buildernames = set(buildernames)
```
armenzg: If "considered_list_of_builders" is not defined, we can call list_builders(repo_name=repo_name)

mozci/sources/buildbot_bridge.py
```diff
@@ -303,6 +308,25 @@ def _generate_tasks(repo_name, revision, builders_graph, task_graph_id=None,
     return tasks
 
 
+def trigger_missing_jobs_for_revision(repo_name, revision, dry_run=False):
```
armenzg: Re-stating what I said above.

Let's move trigger_missing_jobs_for_revision() to TaskClusterBuildbotAPIManager().

armenzg:
Replying to Alice:

> Another problem that I saw, but that I have no idea of what is going on is that we are finding a lot of buildernames that are not of either coalesced jobs or existing jobs: around 700 per push. A lot of Windows XP pgo jobs on inbound, for example. @armenzg do you know what is going on?

From our conversation on IRC, it seems that this is alright.
We could modify list_builders() to only return "opt" jobs if there is both.
Not necessary for this PR though.

armenzg:
Replying to Alice:

> Another problem that I saw, but that I have no idea of what is going on is that we are finding a lot of buildernames that are not of either coalesced jobs or existing jobs: around 700 per push. A lot of Windows XP pgo jobs on inbound, for example. @armenzg do you know what is going on?

From our conversation on IRC, it seems that this is alright.
We could modify list_builders() to only return "opt" jobs if there is both.
Not necessary for this PR though.
I filed this which is not important:
https://github.com/mozilla/mozilla_ci_tools/issues/395

armenzg:
I also assigned you to https://github.com/mozilla/mozilla_ci_tools/issues/392 since you already have the code coming up.

armenzg:
Do you want us to look at the code or still working on it? (I see two more commits 21 hours ago)

F3real:
@armenzg You can review it now.

mozci/query_jobs.py
```diff
@@ -141,6 +142,71 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
...
+                else:
+                    if buildername in considered_list_of_builders:
+                        correct_status_builders.add(buildername)
+            except BuildjsonError:
+                LOG.info('We were not able to find status information for "%s"'
+                         % buildername)
+
+        correct_status_builders = correct_status_builders - wrong_status_builders
+        return list(correct_status_builders)
+
```
adusca: I believe the 3 functions could be QueryApi methods, since they only depend on _get_all_jobs and get_job_status, and both functions are implemented on BuildApi and in TreeherderApi (you don't need to implement the functions on QueryApi to be able to call a function that depends on them, see https://pastebin.mozilla.org/8853248 )

adusca:
Right now there are 2 functions called trigger_missing_jobs_for_revision, one in ci_manager, as a TaskClusterBuildbotManager method and the other in mozci.py. Can you move the one in mozci.py to be under BuildAPIManager in ci_manager?

In the future, the user will select the ci_manager and we just call ci_manager.trigger_missing_jobs_for_revision.

F3real:
@adusca Does this look ok ?

mozci/ci_manager.py
```diff
@@ -153,4 +178,20 @@ def schedule_arbitrary_job(self, repo_name, revision, uuid, *args, **kwargs):
         return super(TaskClusterBuildbotManager, self).schedule_graph(
             task_graph=task_graph, *args, **kwargs)
 
+    def trigger_missing_jobs_for_revision(self, repo_name, revision, dry_run=False):
```
adusca: Both trigger_missing_jobs_for_revision's should have the same API, so we need to add support to trigger_build_if_missing here. We could just receive the parameter and do nothing with it, and open a new issue to add proper support for trigger_build_if_missing here, because the patch already does a lot.

If you want to do it as part of this patch, to add trigger_build_if_missing here we need to pass the argument to buildbot_graph_builder. There we need to add an extra if clause to only add the build job and the test job if trigger_build_if_missing is False (here: https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/sources/buildbot_bridge.py#L166 )

adusca:
This is almost ready, there is just the last comment I mentioned. When you are done with the changes (either by tackling the trigger_build_if_missing part now or by leaving for a follow up issue), please squash your commits so I can merge them.

F3real:
@adusca  rebased and squashed. I just changed function signatures to be same there should be sperate issue to fix that.

adusca:
The code is already on master, so I'm going to close this (already merged) PR. 

What happened: I tried merging from the command-line using cherry-pick and now the commit shows up in: https://github.com/mozilla/mozilla_ci_tools/commits/master (I think the history looks prettier without the merge commits) but GitHub didn't mark this PR as merged automatically after that. If this causes anyone any problem, we can go back to pressing "Merge pull request" on the GitHub interface.

Issue #382, opened 2015-11-20
Do not delete files automatically


armenzg:
It seems that we automatically clean files older than 30 days [1]
This is a very low priority issue.
This was needed for pulse_actions not to store too much data on disk, however, the dynos are restarted every day and we start from a clean slate.

This became a nuissance when I was trying to work on a push older than 30 days.

Let's not delete by default, change the value to 120 days (since TH does not keep more than 4 months of data) and have a global variable if any consumer wants to change the behaviour.

11/20/2015 12:10:47 mozci DEBUG:	 Cleaning up /home/armenzg/.mozilla/mozci/builds-2015-09-15.js

[1] https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/mozci.py#L412

xennygrimmato:
Has this been fixed already? If not, I can take care of this and create a PR.

armenzg:
Assigned! Thanks @xennygrimmato

xennygrimmato:
So, since we do not want to clean by default, should I just remove this line:
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/mozci.py#L412

And insert a global `CLEANUP_DAYS` variable in:
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/utils/transfer.py 

I will modify:
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/utils/transfer.py#L44
with:
```
days=CLEANUP_DAYS
```
Does this look ok to you?

armenzg:
Yes to all :)

PR #389, opened 2015-11-24
Issue 382 - Do not delete files by default & change the default value to 120 days

xennygrimmato:
This is a fix for issue: https://github.com/mozilla/mozilla_ci_tools/issues/382

### DIFF
--- a/mozci/utils/transfer.py
+++ b/mozci/utils/transfer.py
@@ -26,6 +26,7 @@ except:
 LOG = logging.getLogger('mozci')
 MEMORY_SAVING_MODE = False
 SHOW_PROGRESS_BAR = True
+CLEANUP_DAYS = 120
 
 
 def path_to_file(filename):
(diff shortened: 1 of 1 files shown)


mozci/mozci.py
```diff
@@ -409,9 +409,6 @@ def trigger_job(revision, buildername, times=1, files=None, dry_run=False,
     else:
         LOG.debug("Nothing needs to be triggered")
 
-    # Cleanup old buildjson files.
-    clean_directory()
-
```
vaibhavmagarwal, line 414: Hi @xennygrimmato, thanks for working on this! I think we should not remove this otherwise clean_directory will never be called and the files older than 120 days will not be cleaned up. I think we should delete files older than 120 days since Treeherder also does not keep data older than it, so no use keeping the data on disk and occupying space.

xennygrimmato, line 414: Ok! Added it back. `clean_directory()` gets called just like it did, before - only the default value of days is now set to 120.

mozci/mozci.py
```diff
@@ -36,7 +36,7 @@
     TreeherderApi
 )
 from mozci.utils.misc import _all_urls_reachable
-from mozci.utils.transfer import path_to_file, clean_directory
+from mozci.utils.transfer import path_to_file
```
adusca: Looks like you are no longer importing clean_directory() but are still calling it.

I don't think calling clean_directory every time we call trigger_job is a good idea. I believe users either want mozci to clean up itself immediately (so we clean every builds*.js file we download immediately after running the script) or want to keep the files around for as long as they are needed (and so would want to manually remove them).

vaibhavmagarwal: I agree that calling clean_directory every time with trigger_job is not such a good idea. But we should also not clean up the build*.js files immediately, otherwise we would take a lot of resources for downloading every time when we can easily cache (for same commands). Also, always keeping them is not ideal because it is just taking space on the user's machine. So currently, I can think of this way as the only solution close to ideal, but am open to more thoughts. @armenzg thoughts?

armenzg:
Thank you all!

Issue #384, opened 2015-11-20
Make trigger.py to schedule jobs through TaskCluster instead of Buildapi


armenzg:
trigger.py [1] currently schedules jobs through Buildapi instead of TaskCluster.
We can switch this by scheduling through the classes defined in mozci/ci_manager.py
Technologically, we want to move away from using Buildapi directly since it will be deprecated eventually.

I want to allow people to say --scheduling-source 'buildapi' if they want to but default to TaskCluster.

[1] https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/trigger.py
[2] https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/ci_manager.py

MikeLing:
Hi @armenzg , do you want me to work on this? I'm interesting in this, so you could assign it to me if it's OK: )

MikeLing:
Is that the BuildApi you mean in here (https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/trigger.py#L18)? which defined in query_job? I only find this part is related with BuildApi, is that correct?

adusca:
@MikeLing Right now we do all of our trigger requests (trigger_arbitrary_job, make_retrigger_request) using [BuildApi self-serve](https://secure.pub.build.mozilla.org/buildapi/self-serve). 

We would like to use BBB (Buildbot Bridge) instead. [This](https://github.com/mozilla/pulse_actions/blob/master/pulse_actions/handlers/treeherder_runnable.py#L63) is an example of using BBB to schedule a graph of builders.

adusca:
@armenzg Since using BBB requires different credentials, how would current mozci users be able to continue using trigger.py?

armenzg:
I added some info on how to get credentials in here:
http://mozilla-ci-tools.readthedocs.org/en/latest/buildbot_bridge.html

Do you think this is a good start?

adusca:
@armenzg Would this break trigger.py for our users unless they specify --buildapi? Or it is safe to assume that most of our users will have TC credentials?

armenzg:
It will be fine until we make a release.
We should give people heads up and tell them ahead of time that they should get TC credentials.

MikeLing:
I think this issue could be closed for this PR(https://github.com/mozilla/mozilla_ci_tools/pull/390) had been merged :)

PR #390, opened 2015-11-24
Issue 384 - Make trigger.py to schedule jobs through TaskCluster instead of Buildapi

MikeLing:
It's PR for https://github.com/mozilla/mozilla_ci_tools/issues/384, But it's not good to go yet.

The ideal of this PR is:

> get build_graph
> if schedule_way == "taskCluster"(which is default):
>    trigger_job(option.uuid)
> else:
>    use buildapi(just like we used before)

### DIFF
--- a/mozci/ci_manager.py
+++ b/mozci/ci_manager.py
@@ -53,6 +53,11 @@ class BaseCIManager:
     def cancel_all(self, repo_name, revision, *args, **kwargs):
         pass
 
+    @abstractmethod
+    def trigger_range(self, buildername, repo_name, revisions, times, dry_run, files,
+                      trigger_build_if_missing):
+        pass
+
 # End of BaseCIManager
 
 
@@ -98,6 +103,7 @@ class BuildAPIManager(BaseCIManager):
         for buildername in builders_for_repo:
             trigger_range(
                 buildername=buildername,
+                repo_name=repo_name,
                 revisions=[revision],
                 times=1,
                 dry_run=dry_run,
@@ -108,6 +114,18 @@ class BuildAPIManager(BaseCIManager):
                 },
                 trigger_build_if_missing=trigger_build_if_missing
             )
+
+    def trigger_range(self, buildername, repo_name, revisions, times, dry_run, files,
+                      trigger_build_if_missing):
+        trigger_range(
+            buildername=buildername,
+            repo_name=repo_name,
+            revisions=revisions,
+            times=times,
+            dry_run=dry_run,
+            files=files,
+            trigger_build_if_missing=trigger_build_if_missing
+        )
 
 # End of BuildAPIManager
 
@@ -197,4 +215,18 @@ class TaskClusterBuildbotManager(TaskClusterManager):
             builders_graph=buildbot_graph
             )
 
+    def trigger_range(self, buildername, repo_name, revisions, times, dry_run, files,
+                      trigger_build_if_missing):
+        for revision in revisions:
+            builder_graph, trigger_with_buildapi = buildbot_bridge.buildbot_graph_builder(
+                builders=[buildername],
+                revision=revision,
+                complete=False  # XXX: This can be removed when BBB is in use
+            )
+            self.schedule_graph(
+                repo_name=repo_name,
+                revision=revision,
+                builders_graph=builder_graph,
+                dry_run=dry_run
+            )
 # End of TaskClusterBuildbotManager
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -438,10 +438,10 @@ def trigger_job(revision, buildername, times=1, files=None, dry_run=False,
     return list_of_requests
 
 
-def trigger_range(buildername, revisions, times=1, dry_run=False,
+def trigger_range(buildername, repo_name, revisions, times=1, dry_run=False,
                   files=None, extra_properties=None, trigger_build_if_missing=True):
     """Schedule the job named "buildername" ("times" times) in every revision on 'revisions'."""
-    repo_name = query_repo_name_from_buildername(buildername)
+    assert repo_name in buildername
     repo_url = repositories.query_repo_url(repo_name)
 
     if revisions != []:
--- a/mozci/scripts/trigger.py
+++ b/mozci/scripts/trigger.py
@@ -6,14 +6,13 @@ from argparse import ArgumentParser
 
 from buildapi_client import make_retrigger_request
 
-from mozci.ci_manager import BuildAPIManager
+from mozci.ci_manager import BuildAPIManager, TaskClusterBuildbotManager
 from mozci.mozci import (
     find_backfill_revlist,
     query_builders,
     query_repo_name_from_buildername,
     query_repo_url_from_buildername,
-    set_query_source,
-    trigger_range
+    set_query_source
 )
 from mozci.query_jobs import BuildApi, COALESCED
 from mozci.repositories import query_repo_url
@@ -150,6 +149,10 @@ def parse_args(argv=None):
                         action="store_false",
                         dest="existing_only",
                         help="Only trigger test job if the build jobs already exists.")
+
+    parser.add_argument("--taskcluster",
+                        action="store_true",
+                        help="Schedule jobs through TaskCluster.")
 
     options = parser.parse_args(argv)
     return options
@@ -301,6 +304,13 @@ def main():
     else:
         revision = query_push_by_revision(repo_url, options.rev,
                                           return_revision_list=True)
+
+    # Schedule jobs through TaskCluster if --taskcluster option has been set to true
+    if options.taskcluster:
+        mgr = TaskClusterBuildbotManager()
+    else:
+        mgr = BuildAPIManager()
+
     # Mode 1: Trigger coalesced jobs
     if options.coalesced:
         query_api = BuildApi()
@@ -318,7 +328,7 @@ def main():
 
     # Mode #2: Fill-in a revision or trigger_test_jobs_only
     if options.fill_revision or options.trigger_tests_only:
-        BuildAPIManager().trigger_missing_jobs_for_revision(
+        mgr.trigger_missing_jobs_for_revision(
             repo_name=repo_name,
             revision=revision,
             dry_run=options.dry_run,
@@ -383,8 +393,9 @@ def main():
             exclude=options.exclude)
 
         try:
-            trigger_range(
+            mgr.trigger_range(
                 buildername=buildername,
+                repo_name=repo_name,
                 revisions=revlist,
                 times=options.times,
                 dry_run=options.dry_run,
--- a/mozci/sources/buildbot_bridge.py
+++ b/mozci/sources/buildbot_bridge.py
@@ -135,7 +135,11 @@ def buildbot_graph_builder(builders, revision, complete=True):
            https://github.com/mozilla/mozilla_ci_tools/issues/353
 
     Input: a list of builders and a revision
-    Output: a graph with the builders we received and the necessary upstream jobs
+    Output: a set which includes a graph with the builders we received, the necessary upstream
+            jobs and list of builders which can been used to schedule jobs through trigger_job()
+
+    NOTE: We will make it only return builder graph once Buildbot jobs are scheduled via
+          the Buildbot Bridge (BBB) instead of the normal Buildbot scheduling.
 
     Graph of N levels:
         {
@@ -154,6 +158,8 @@ def buildbot_graph_builder(builders, revision, complete=True):
     :type revision: str
     :return: A graph of buildernames (single level of graphs)
     :rtype: dict
+    :param complete: indicate that build has been completed or not
+    :type: booleans
 
     """
     graph = {}
@@ -174,7 +180,9 @@ def buildbot_graph_builder(builders, revision, complete=True):
 
             # The build job is already completed, we can trigger the test job
             if objective == b:
-                # If complete is True, we add the job to our graph
+                # XXX: Fix me - Adding test jobs to the graph without a build associated
+                # to it does not work. This will be fixed once we switch to scheduling
+                # Buildbot jobs through BBB
                 if complete:
                     graph[b] = None
                 else:
@@ -377,11 +385,11 @@ def generate_builders_tc_graph(repo_name, revision, builders_graph, *args, **kwa
     """
     if builders_graph is None:
         return None
-
-    metadata = kwargs.get(
-        'metadata',
-        generate_metadata(repo_name=repo_name, revision=revision, name='Mozci BBB graph')
-    )
+    metadata = kwargs.get('metadata')
+    if metadata is None:
+        metadata = generate_metadata(repo_name=repo_name,
+                                     revision=revision,
+                                     name='Mozci BBB graph')
     # This is the initial task graph which we're defining
     task_graph = generate_task_graph(
         scopes=[
--- a/mozci/sources/tc.py
+++ b/mozci/sources/tc.py
@@ -66,8 +66,8 @@ def generate_metadata(repo_name, revision, name,
     :type description: str
     """
     repo_url = query_repo_url(repo_name)
-    push_info = push_info = query_push_by_revision(repo_url=repo_url,
-                                                   revision=revision)
+    push_info = query_push_by_revision(repo_url=repo_url,
+                                       revision=revision)
 
     return {
         'name': name,


MikeLing:
Hey @adusca and @armenzg ,

Thank you for all your help and comment :)

I make a few progress for this, but I'm not sure if I'm not the right direction(actually, I'm quite unsure). Could you give me some hit about what should I do next? Thanks a lot! :)

mozci/scripts/trigger.py
```diff
@@ -156,6 +167,13 @@ def validate_options(options):
         if options.from_rev:
             error_message = "You should not pass --end-rev " \
                             "when you use --delta."
+    if options.schedule_way:
+        if options.schedule_way != "buildapi" and options.schedule_way != "taskCluster":
+            error_message = "We don'y have other schedule job way except 'buildapi' and 'TaskCluster' " \
+                            "and the default way is TaskCluster."
+    if options.schedule_way == "taskCluster":
+        if not options.task_id:
```
adusca: Why ask for task id? I though we would create a new task.

mozci/scripts/trigger.py
```diff
@@ -245,61 +264,71 @@ def main():
...
-                                   dry_run=options.dry_run)
-
-        return
-
-    # Mode #2: Fill-in a revision
-    if options.fill_revision:
-        trigger_missing_jobs_for_revision(
+    if options.schedule_way == "taskCluster":
+        builders_graph = buildbot_bridge.buildbot_graph_builder(options.buildernames, revision)
+        mgr.schedule_graph(
```
adusca: Right now, the only mode for which both BBB and buildapi are supported is triggering a list of buildernames in just one revision, so only this mode should be in the if/else block

mozci/scripts/trigger.py
```diff
@@ -128,6 +130,15 @@ def parse_args(argv=None):
                         help="We will trigger jobs starting from --rev in reverse chronological "
                         "order until we find the last revision where there was a good job.")
 
+    parser.add_argument("--scheduling-source",
```
adusca: --scheduling-source buildapi is a bit too verbose. Since there is only 2 options and one is going to be the default, we could do --buildapi instead.

mozci/scripts/trigger.py
```diff
@@ -128,6 +130,15 @@ def parse_args(argv=None):
...
+                        dest="schedule_way",
+                        htlp="We could switch the way we schedule job from TaskCluster "
```
adusca: help=Schedule jobs through BuildAPI

armenzg:
Looking at this now.

mozci/scripts/trigger.py
```diff
@@ -23,8 +23,10 @@
     query_revisions_range_from_revision_before_and_after,
     query_full_revision_info
 )
+from mozci.sources import buildbot_bridge
 from mozci.utils.authentication import valid_credentials
 from mozci.utils.log_util import setup_logging
+from mozci.ci_manager import TaskclusterManager
```
armenzg: Could you please use TaskClusterBuildbotManager instead?
We don't yet have the capacity to work completely with TaskCluster pure tasks. We only can handle tasks which would trigger Buildbot jobs (via Buildbot Bridge).

If you haven't, could you please read this blog post?
http://armenzg.blogspot.ca/2015/09/mozilla-ci-tools-upcoming-support-for.html
http://hearsum.ca/blog/buildbot-taskcluster-bridge-an-overview.html
It should help with understanding a bit more what the Buildbot Bridge (BBB) is and how it works.

I will be adding more comments after this one.

mozci/scripts/trigger.py
armenzg: A bit too verbose as in a long option name?

I'm happy with --buildapi since TC would be the default. Good idea!

mozci/scripts/trigger.py
```diff
@@ -128,6 +130,15 @@ def parse_args(argv=None):
...
                         "order until we find the last revision where there was a good job.")
 
+    parser.add_argument("--scheduling-source",
+                        dest="schedule_way",
+                        htlp="We could switch the way we schedule job from TaskCluster "
+                             "to Buildapi by --scheduling-source buildapi")
+
+    parser.add_argument("--task-id",
+                        dest="task_id",
+                        help="The the taskId we used to referencing required tasks.")
```
armenzg: We don't need a task id.
We're going to be focusing on Buildbot jobs which we can schedule through TaskCluster instead of Buildapi.

This means that we can refer them with the buildername.

mozci/scripts/trigger.py
armenzg: I answered above. We don't need this for tasks which trigger Buildbot builders.

mozci/scripts/trigger.py
```diff
@@ -220,6 +238,7 @@ def main():
     options = parse_args()
     validate_options(options)
     repo_url = query_repo_url(options.repo_name)
+    mgr = TaskclusterManager()
```
armenzg: Use the other manager I mentioned above.

armenzg:
I think we need to split up this work into various pieces to make this work easier.

Let's start by still defaulting to buildapi instead of taskcluster.
Let's first focus on making indrect calls to functions from mozci.py.

Currently, we make calls to mozci.py like this:
- make_retrigger_request
- trigger_missing_jobs_for_revision
- trigger_range

What I would like to see is add some equivalent functions to BaseCIManager and be implemented by the subclasses.

This means that we would do something like this:
- make_retrigger_request -> mgr.retrigger()
  - Forcefully use a BuildAPIManager as we can't do this through TC
  - For TC we will need to take a list of jobs that were coalesced an schedule as a whole graph
- trigger_missing_jobs_for_revision -> mgr.trigger_missing_jobs()
- trigger_range -> mgr.trigger_range()

Do not move the functions out of mozci.py. Just call them from ci_manager.py. We're simply adding indirection.

For TC, to have trigger_missing_jobs, we would have to separate it into two stages:
- list of jobs = determine_missing_jobs()
- schedulde_graph( build a graph based on list of jobs)

For trigger_range, we would:
- list of revisions = determine_trigger_range()
- schedule a graph per revision

I think determine_missing_jobs and determine_trigger_range could be implemented functions of BuildAPIManager which the subclasses call.

Could you please make these changes and request for another review?

That way we can first land a piece of code and then continue with follow up changes.

armenzg:
I'm going to split this work up to allow make smaller improvements and some of it already xenny was looking into.

MikeLing:
hello @armenzg,

So, which part is my job to do?

armenzg:
My apologies @MikeLing. The last comment must have been confusing!

Could you please make trigger.py use BuildAPIManager and TaskClusterBuildbotManager? (depending if the user chooses --taskcluster).

We can't yet switch to TaskCluster as the default scheduling agent because we don't have trigger_missing_jobs_for_revision and trigger_range.

For now let's focus on building the skeleton for when we have the pieces.

adusca:
Hi! I added the label "needs code change" to your PR. When you are ready for a review, please leave a comment with "r?" and we will look at it.

If you want a specific reviewer, leave a comment with r? @-mention

MikeLing:
np, I will finish it within today :)

mozci/scripts/trigger.py
```diff
@@ -260,6 +268,20 @@ def main():
 
     else:
         revision = query_full_revision_info(repo_url, options.rev)
+
+    # Schedule jobs through TaskCluster if --taskcluster option been set as true
+    if options.taskcluster:
+        mgr = TaskClusterBuildbotManager()
+        # Trigger missing jobs for a given revision.
```
armenzg: Could you please change the comment to "# Mode #2: Fill-in a revision or trigger_test_jobs_only"?

mozci/scripts/trigger.py
```diff
@@ -260,6 +268,20 @@ def main():
...
+        if options.fill_revision:
+            mgr.trigger_missing_jobs_for_revision(options.repo_name, revision)
```
armenzg: Could you please add "return" after this line?
Look at this line:
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/trigger.py#L285

mozci/scripts/trigger.py
```diff
@@ -260,6 +268,20 @@ def main():
...
+
+        builder_graph = buildbot_bridge.buildbot_graph_builder(options.buildernames, revision)
```
armenzg: Could you please put this inside of a "if options.buildernames" block?
Also add the same note as in here:
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/trigger.py#L287

armenzg:
Hi @MikeLing this is going in the right direction.
I added some minor comments above.

Could you also do this?
1. add a trigger_range function in this class [1]
2. make that call mozci.py's trigger_range()
3. then, make this trigger_range() [2] call BuildAPIManager's trigger_range() (added on the first step)

[1] https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/ci_manager.py#L58
[2] https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/trigger.py#L301

armenzg:
@adusca FYI Mikeling asked for feedback through IRC. That is why he did not use the "r?" comment in here.

MikeLing:
I will address it and ask review again: )

MikeLing:
@armenzg r?
BTW, maybe we need some more tests for it? I guess I need to add those tests in test_allthethings.py file.

mozci/scripts/trigger.py
```diff
@@ -260,6 +267,21 @@ def main():
 
     else:
         revision = query_full_revision_info(repo_url, options.rev)
+
+    # Schedule jobs through TaskCluster if --taskcluster option been set as true
```
armenzg: "option been set as true" -> "option has been set to true"

mozci/scripts/trigger.py
```diff
@@ -260,6 +267,21 @@ def main():
...
+    if options.taskcluster:
+        mgr = TaskClusterBuildbotManager()
```
armenzg, line 310:
```
if options.taskcluster:
    mgr = TaskClusterBuildbotManager()
else:
    mgr = BuildAPIManager()
```

Doing that we won't need to do the next 6 lines. More comments below.

mozci/scripts/trigger.py
```diff
@@ -260,6 +267,21 @@ def main():
...
+        builder_graph = buildbot_bridge.buildbot_graph_builder(options.buildernames, revision)
+        # Mode #2: Fill-in a revision or trigger_test_jobs_only
+        if options.fill_revision:
+            mgr.trigger_missing_jobs_for_revision(options.repo_name, revision)
+            return
+        # Mode #3: Trigger jobs based on revision list modifiers
```
armenzg: Can you please move this logic into TaskClusterBuildbotManager as a function called trigger_range()?

```
builder_graph = buildbot_bridge.buildbot_graph_builder(options.buildernames, revision)
mgr.schedule_graph(...)
```

We should have a loop over the list of revisions and schedule a graph per revision.
Please use the same signature as the trigger_range you added for BuildAPIManager

Could you also please add to the Base class this function?

```
 def trigger_range(self, buildername, revisions, dry_run=False, trigger_build_if_missing=True, *args, **kwargs):
```

mozci/scripts/trigger.py
```diff
@@ -298,7 +320,7 @@ def main():
             max_revisions=options.max_revisions)
 
         try:
-            trigger_range(
+            BuildAPIManager().trigger_range(
```
armenzg: Use in here mgr.trigger_range()

armenzg:
This is looking good.
We need to do a bit of refactoring but it is going well.
Thanks Mike!

MikeLing:
@armenzg do we need further address for it or it's good to go? r?

mozci/scripts/trigger.py
```diff
@@ -298,7 +311,8 @@ def main():
             max_revisions=options.max_revisions)
 
         try:
-            trigger_range(
+            mgr.trigger_range(
+                repo_name=options.repo_name,
```
adusca: BuildAPI's trigger_range doesn't take repo_name as a parameter, this gives us the following (very confusing) error:

```
Traceback (most recent call last):
  File "mozci/scripts/trigger.py", line 321, in main
    trigger_build_if_missing=options.trigger_build_if_missing
TypeError: trigger_range() takes at least 3 arguments (4 given)
```

It is very important that the methods have the same signature in both classes.

adusca:
Testing this patch I discovered that trigger.py has been broken for a while. We don't require users to pass a repo-name when they pass a buildername (because we can get the repo-name from the buildername) but we are doing:

```
    repo_url = query_repo_url(options.repo_name)
```

in the beginning of main(), which gives us the following error when users don't pass --repo-name:

```
(venv)mozilla_ci_tools[MikeLing-issuefix-384 $]$ python mozci/scripts/trigger.py -r 71cf9b8cc82e --dry-run --debug -b "Windows 7 32-bit try opt test mochitest-1"
Traceback (most recent call last):
  File "mozci/scripts/trigger.py", line 335, in <module>
    main()
  File "mozci/scripts/trigger.py", line 243, in main
    repo_url = query_repo_url(options.repo_name)
  File "/home/adusca/mozilla_ci_tools/mozci/repositories.py", line 41, in query_repo_url
    return query_repository(repo_name)["repo"]
  File "/home/adusca/mozilla_ci_tools/mozci/repositories.py", line 34, in query_repository
    raise MozciError("That repository does not exist.")
mozci.errors.MozciError: That repository does not exist.
```

A repo-name should not be mandatory when users pass a buildername, and even if we decided to make it mandatory there should be an error message about it.

mozci/scripts/trigger.py
adusca: If BBB's trigger_range really needs to receive a repo_name, we can determine the repo_name from the buildername and pass it to all versions of trigger_range.

MikeLing: Hmm, how about add all repo_name to all trigger_range function but we don't use it in BuildAPIManager's trigger_range?  I'm not sure if it's a good programming habit or not(pass a variable but not use it)

armenzg:
Hi @MikeLing , any more changes before we look at this?

MikeLing:
@armenzg , yep. I will ask review after I think it's good to go :)

MikeLing:
Hmm, why it keeps telling me **TypeError: 'list' object is not callable**. 

> python mozci/scripts/trigger.py --repo-name try --revision a1e682c00ff3 -b "Ubuntu VM 12.04 x64 try debug test mochitest-4" --dry-run --debug
> I had reset to the early version but the error still shows up.

armenzg:
Hi @MikeLing any progress in here?

mozci/mozci.py
```diff
@@ -438,10 +438,9 @@ def trigger_job(revision, buildername, times=1, files=None, dry_run=False,
     return list_of_requests
 
 
-def trigger_range(buildername, revisions, times=1, dry_run=False,
+def trigger_range(buildername, repo_name, revisions, times=1, dry_run=False,
                   files=None, extra_properties=None, trigger_build_if_missing=True):
     """Schedule the job named "buildername" ("times" times) in every revision on 'revisions'."""
-    repo_name = query_repo_name_from_buildername(buildername)
```
MikeLing, line 444: I think we no longer need to query repo name from buildername, because we had do it in https://github.com/mozilla/mozilla_ci_tools/pull/390/files#diff-14a3d986e72396dbc0d1a2774fee1046R294.

armenzg, line 444: We would then need to assert that repo_name is contained in buildername just to be safe.
The script is not the only tool that could be calling trigger_range().

mozci/ci_manager.py
```diff
@@ -109,6 +110,18 @@ def trigger_missing_jobs_for_revision(self, repo_name, revision, dry_run=False,
                 trigger_build_if_missing=trigger_build_if_missing
             )
 
+    def trigger_range(self, buildername, repo_name, revisions, times, dry_run, files,
+                      trigger_build_if_missing):
+        trigger_range(
```
armenzg, line 120: Does this not become a recursive call?
How should it know that it is supposed to call the trigger_range from the mozci module instead of itself's?

armenzg, line 120: Oh right. Instead of self.trigger_range()
Nevermind :)

mozci/ci_manager.py
```diff
@@ -197,4 +210,17 @@ def trigger_missing_jobs_for_revision(self, repo_name, revision, dry_run=False,
             builders_graph=buildbot_graph
             )
 
+    def trigger_range(self, buildername, repo_name, revisions, times, dry_run, files,
+                      trigger_build_if_missing):
+        for revision in revisions:
+            builder_graph = buildbot_bridge.buildbot_graph_builder(
```
armenzg: Could you do:
builder_graph, trigger_with_buildapi = ...
and remove the [0]. Also use complete=False
Here's a code sample:
https://github.com/mozilla/pulse_actions/blob/master/pulse_actions/handlers/treeherder_runnable.py#L88

Some test jobs _cannot_ be scheduled directly via TaskCluster/BBB since there are some limitations on the Buildbot side.
In that case, trigger_with_buildapi builders should be scheduled through trigger_job()
Could you also add a note about this? That we will fix it once Buildbot jobs are scheduled via the Buildbot Bridge (BBB) instead of the normal Buildbot scheduling?

Could you also update the string doc for buildbot_graph_builder() to indicate what to expect as return values?
It got me confused. I should have asked it to be added when landed.

Could you also add a note about the variable 'complete'?

Also change the note of:
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/sources/buildbot_bridge.py#L177
to:

>  XXX: Fix me - Adding test jobs to the graph without a build associated to it does not work
>           This will be fixed once we switch to scheduling Buildbot jobs through BBB

mozci/scripts/trigger.py
```diff
@@ -151,6 +150,10 @@ def parse_args(argv=None):
                         dest="existing_only",
                         help="Only trigger test job if the build jobs already exists.")
 
+    parser.add_argument("--taskcluster",
```
armenzg, line 153: Good!

armenzg:
Wonderful!
This is looking good.

NOTE to self: In the next pass I will need to checkout the code and try it with few different options.

mozci/mozci.py
MikeLing, line 444: Hmm, then we should require user query accurate repo name by (e.g query_repo_name_from_buildername). I think separate our function call can make sure we have a clearer logic. :)

But I totally defer to your idea about it, I will change it back if you think we need query repo inside.

mozci/ci_manager.py
MikeLing, line 120: oh, yeah, I should add trigger_range into the Base Class also :)

mozci/ci_manager.py
MikeLing: Done :)

MikeLing:
@armenzg r ?

mozci/mozci.py
armenzg, line 444: It's fine to add repo_name but we need to make to assert that repo_name is contained within buildername.
Otherwise, we could try to schedule a buildername that belongs to a different repo_name.

Could you please add the assert?

mozci/sources/buildbot_bridge.py
```diff
@@ -135,7 +135,11 @@ def buildbot_graph_builder(builders, revision, complete=True):
            https://github.com/mozilla/mozilla_ci_tools/issues/353
 
     Input: a list of builders and a revision
-    Output: a graph with the builders we received and the necessary upstream jobs
+    Output: a set which include a graph with the builders we received and the necessary upstream
```
armenzg: include --> plural
"received and the" --> "received, the" (notice the comma)

mozci/sources/buildbot_bridge.py
```diff
@@ -135,7 +135,11 @@ def buildbot_graph_builder(builders, revision, complete=True):
...
+            jobs and list of builders can been used to schedule jobs through trigger_job()
```
armenzg: "builders can" --> "builders which can"

armenzg:
This is looking great.

I noticed a bug that I have not yet investigated.
Would you mind trying to figure out why 'metadata' is null?

armenzg@armenzg-thinkpad:~/repos/ci_tools$ python mozci/scripts/trigger.py --query-source treeherder --taskcluster --repo-name try --revision cd403eeef6ad --times 5 --buildername 'Android 4.3 armv7 API 11+ try debug test plain-reftest-41' --dry-run
01/22/2016 11:41:02 root INFO:  Setting INFO level
01/22/2016 11:41:05 root INFO:  https://treeherder.mozilla.org/#/jobs?repo=try&revision=cd403eeef6ad7e391c22a3910f99c98ca03c4c0a
01/22/2016 11:41:10 mozci INFO: Outputting the graph:
{
    "scopes": [
        "queue:define-task:buildbot-bridge/buildbot-bridge", 
        "scheduler:create-task-graph"
    ], 
    "tasks": [
        {
            "reruns": 0, 
            "taskId": "LUPkO7WgSQ6c4Xn8WwSUNg", 
            "task": {
                "scopes": [], 
                "extra": {}, 
                "expires": "2016-01-23T16:41:09.228517Z", 
                "schedulerId": "task-graph-scheduler", 
                "deadline": "2016-01-23T16:41:09.228501Z", 
                "payload": {
                    "sourcestamp": {
                        "branch": "try", 
                        "revision": "cd403eeef6ad7e391c22a3910f99c98ca03c4c0a"
                    }, 
                    "properties": {
                        "upload_to_task_id": "D6cB2a4bRn-qu4pETBNshQ", 
                        "product": "mobile", 
                        "who": "armenzg@mozilla.com"
                    }, 
                    "buildername": "Android armv7 API 11+ try debug build"
                }, 
                "provisionerId": "buildbot-bridge", 
                "workerType": "buildbot-bridge", 
                "retries": 5, 
                "created": "2016-01-22T16:41:09.228465Z", 
                "tags": {}, 
                "priority": "normal", 
                "routes": [], 
                "metadata": {
                    "owner": "armenzg@mozilla.com", 
                    "source": "https://hg.mozilla.org/try/rev/cd403eeef6ad7e391c22a3910f99c98ca03c4c0a", 
                    "name": "Android armv7 API 11+ try debug build", 
                    "description": "Task graph generated via Mozilla CI tools"
                }
            }
        }, 
        {
            "reruns": 0, 
            "requires": [
                "LUPkO7WgSQ6c4Xn8WwSUNg"
            ], 
            "taskId": "O1VtufZhR-OcxIZI8G_Whw", 
            "task": {
                "scopes": [], 
                "extra": {}, 
                "expires": "2016-01-23T16:41:10.942868Z", 
                "schedulerId": "task-graph-scheduler", 
                "deadline": "2016-01-23T16:41:10.942853Z", 
                "payload": {
                    "sourcestamp": {
                        "branch": "try", 
                        "revision": "cd403eeef6ad7e391c22a3910f99c98ca03c4c0a"
                    }, 
                    "properties": {
                        "upload_to_task_id": "UvBlKYKmR4uNJNGYgLwm5A", 
                        "product": "mobile", 
                        "who": "armenzg@mozilla.com", 
                        "parent_task_id": "D6cB2a4bRn-qu4pETBNshQ"
                    }, 
                    "buildername": "Android 4.3 armv7 API 11+ try debug test plain-reftest-41"
                }, 
                "provisionerId": "buildbot-bridge", 
                "workerType": "buildbot-bridge", 
                "retries": 5, 
                "created": "2016-01-22T16:41:10.942824Z", 
                "tags": {}, 
                "priority": "normal", 
                "routes": [], 
                "metadata": {
                    "owner": "armenzg@mozilla.com", 
                    "source": "https://hg.mozilla.org/try/rev/cd403eeef6ad7e391c22a3910f99c98ca03c4c0a", 
                    "name": "Android 4.3 armv7 API 11+ try debug test plain-reftest-41", 
                    "description": "Task graph generated via Mozilla CI tools"
                }
            }
        }
    ], 
    "metadata": null
}
01/22/2016 11:41:10 mozci INFO: DRY-RUN: We have not scheduled the graph.

mozci/sources/buildbot_bridge.py
```diff
@@ -377,11 +385,11 @@ def generate_builders_tc_graph(repo_name, revision, builders_graph, *args, **kwa
...
-
-    metadata = kwargs.get(
-        'metadata',
-        generate_metadata(repo_name=repo_name, revision=revision, name='Mozci BBB graph')
-    )
+    metadata = kwargs.get('metadata')
+    if metadata is None:
+        metadata = generate_metadata(repo_name=repo_name,
+                                     revision=revision,
+                                     name='Mozci BBB graph')
```
armenzg, line 392: Thanks for fixing this!

armenzg:
Once you have the assert comment addressed feel free to merge.

Thanks for your help!

armenzg:
Sorry. I closed it by mistake :)

PR #394, opened 2015-11-25
Issue 326 - Added new flag --trigger-only-test-jobs

F3real:


mozci/scripts/trigger.py
```diff
@@ -128,6 +128,11 @@ def parse_args(argv=None):
                         help="We will trigger jobs starting from --rev in reverse chronological "
                         "order until we find the last revision where there was a good job.")
 
+    parser.add_argument("--trigger-only-test-jobs",
+                        action="store_true",
+                        dest="trigger_test_only",
```
armenzg: Could you please pluralize the word test?
trigger_tests_only

mozci/scripts/trigger.py
```diff
@@ -137,9 +142,10 @@ def validate_options(options):
     Raises an exception if options are missing or conflicting.
     """
     error_message = ""
-    if not options.buildernames and (not options.coalesced and not options.fill_revision):
-        error_message = "A buildername is mandatory for all modes except --coalesced and " \
-                        "--fill-revision. Use --buildername."
+    if (not options.buildernames and
+       (not options.coalesced and not options.fill_revision and not options.trigger_test_only)):
```
armenzg: We want to allow not needing to specify "buildernames" for --trigger-tests-only"
See the developers' command on issue 326

armenzg:
Could you also change the flag --existing-only to --trigger-build-if-missing?
It is more obvious (even if it is a bit harder to type)

This is looking good. After you make those changes could you please squash your changes?
http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

I would merge it after that.

mozci/scripts/trigger.py
armenzg: Nevermind this. Double negations are hard!

armenzg:
Thank you!

PR #399, opened 2015-12-02
Move misc scripts to misc_scripts/

armenzg:
@adusca what do you think?
I think coverage might increase with this change and will help telling people to ignore it.


adusca:
r+

I don't know if someone is using this, but you will need to update the cron job that generates http://people.mozilla.org/~armenzg/permanent/results_per_suite_try.txt . I remember an old IRC conversation in which you told me that generate_list_of_testers_without_builders.py was being used in buildbot, but I can't find that conversation now

adusca:
Found the comment! http://logs.glob.uno/?c=mozilla%23ateam&s=1+Mar+2015&e=29+Mar+2015#c919965

The code in https://github.com/mozilla/build-buildbot-configs/commit/f60361d2e1a989e38d2ba2f2b5fa78f7378551e7 doesn't call the script directly, so nothing needs to be done :)

Issue #366, opened 2015-10-21
Give SETA awareness to mozci


armenzg:
Whenever we fetch a new allthethings.json we should also determine the schedule at which SETA runs.
This is useful for automatic backfilling since we want to know how far we need to go to find an issue.

For instance, our current SETA skipping level is of 7 pushes instead of 5 pushes [1]
The information is contained within the scheduler's name:
    "tests-fx-team-snowleopard-opt-unittest-7-3600":

For now, the same value is used across trees.

[1] https://github.com/mozilla/pulse_actions/blob/master/pulse_actions/handlers/backfilling.py#L25

F3real:
I am interested in working on this.

PR #400, opened 2015-12-02
Add SETA awarness to mozci #366

F3real:


armenzg:
This is looking good.
We need a global dict which if not set we calculate the data.
If it's set, we then return the global dict.

We also need a function which given a buildername, we return the number of pushes for it.

For instance, in pulse_actions, we're using a MAX_REVISIONS variable [1] for every builder, however, we know that Windows XP builders have a different cadence.

A function probably called "get_SETA_info()".

[1] https://github.com/mozilla/pulse_actions/blob/master/pulse_actions/handlers/backfilling.py#L52

F3real:
@armenzg r?

mozci/platforms.py
```diff
@@ -94,6 +95,41 @@ def _process_data():
             BUILDERNAME_TO_TRIGGER[buildername.lower()] = values['triggered_by'][0]
 
 
+def get_SETA_interval_dict(force=False):
+
```
armenzg, line 110: Could you please add a docstring?

mozci/platforms.py
```diff
@@ -94,6 +95,41 @@ def _process_data():
...
+        if not sched.startswith('tests-'):
+            continue
+
+        sched_str_list = sched.split('-')
+        try:
+            for buildername in values['downstream']:
+                seta = []
+                seta.append(int(sched_str_list[-2]))
+                seta.append(int(sched_str_list[-1]))
+                SETA_DICT[buildername] = seta
```
armenzg, line 132: Could you please add comments to explain what are we storing in here? (e.g. push and interval in minutes)

mozci/platforms.py
```diff
@@ -94,6 +95,41 @@ def _process_data():
...
+        except (ValueError, IndexError):
+            continue
+    return SETA_DICT
+
+
+def get_SETA_info(buildername):
```
armenzg, line 139: Docstring and explain what we return.

mozci/platforms.py
```diff
@@ -94,6 +95,41 @@ def _process_data():
...
+        sched_str_list = sched.split('-')
+        try:
+            for buildername in values['downstream']:
+                seta = []
+                seta.append(int(sched_str_list[-2]))
+                seta.append(int(sched_str_list[-1]))
+                SETA_DICT[buildername] = seta
+        except (ValueError, IndexError):
+            continue
+    return SETA_DICT
```
armenzg, line 136: A line blank above the return (visually  more clear).

armenzg:
This looks good! Well done.
Once addressed the nits I can merge it and land it.

Once this is released we can change pulse_actions to use this (not that we to this in this quarter).

PR #401, opened 2015-12-02
Separate modules into the CI they're related to

armenzg:
This is a disruptive change as current PRs will have to rebase if this lands first.
I would like to coordinate this with our contributors.

I believe the benefits of this change is that it will make:
- common modules be at the root
- each CI and modules separated
- pushlog moved as a utils module

This change is for long term benefit.


adusca:
We have 3 schedulers types (BBB, BuildAPI, pure TC) and 2 underlining build systems: Buildbot and TC. BBB schedules Buildbot jobs on TC, so should it be classified under Buildbot or TC? Or should we just divide mozci according to the schedulers? platforms.py is used by both BBB and BuildAPI to determine what to trigger. We can divide things between TC and Buildbot and have BBB and BuildAPI as subdivisions of Buildbot. So many questions!

I agree that we need to organize better how mozci is divided, but I don't now what is the better way.

armenzg:
I will get to this PR after Mozlando.

PR #402, opened 2015-12-04
issue 383 - Merge triggerbyfilters.py with trigger.py

MikeLing:
After move triggerbyfilters into trigger.py, we should use trigger.py like blow:
if you want trigger build normal, you should add a **trigger** command after trigger.py like:
 python mozci/scripts/trigger.py trigger --repo-name try --revision a1e682c00ff3 -b "Ubuntu VM 12.04 x64 try debug test mochitest-4" --dry-run --debug

if you want trigger build by filter, you should add a **filter** command after trigger.py [haven't test yet]


MikeLing:
f? @adusca @armenzg

mozci/scripts/trigger.py
```diff
@@ -25,115 +25,163 @@
 )
 from mozci.utils.authentication import valid_credentials
 from mozci.utils.log_util import setup_logging
+from mozci.platforms import filter_buildernames
 
 
 def parse_args(argv=None):
     """Parse command line options."""
     parser = ArgumentParser()
+    subparsers = parser.add_subparsers()
```
adusca: We don't need 2 subparsers. A lot of our options are common to both scripts and we shouldn't be duplicating so much code.

mozci/scripts/trigger.py
```diff
@@ -231,7 +279,7 @@ def determine_revlist(repo_url, buildername, rev, back_revisions,
     return revlist
 
 
-def main():
+def do_trigger(options):
```
adusca: You are receiving 'options' as an argument and then you do options = parse_args()

mozci/scripts/trigger.py
```diff
@@ -318,5 +366,72 @@ def main():
...
+
+    if options.exclude:
+        filters_out = options.exclude.split(',')
+
+    buildernames = filter_buildernames(
+        buildernames=query_builders(repo_name=options.repo),
+        include=filters_in,
+        exclude=filters_out
+    )
+
```
adusca: There is a lot of code duplication between do_filter and do_trigger. We want "filter" to be an extra mode of main(), that can be identified by the presence of --includes.

mozci/scripts/trigger.py
```diff
@@ -231,7 +279,7 @@ def determine_revlist(repo_url, buildername, rev, back_revisions,
     return revlist
 
 
-def main():
+def do_trigger(options):
     options = parse_args()
     validate_options(options)
```
adusca: You will need to add a new clause to validate_options() to also allow empty buildernames when includes is not empty, in this case, a repo-name should be required. (like we do with coalesced).

adusca:
This is a good start! Right now this patch moves the code from triggerbyfilters.py to trigger.py and renames part of the variables, adding an extra parser. Since 80% of the code is shared between triggerbyfilters.py and trigger.py, we still need to remove this duplication. All of the includes/excludes functionality in triggerbyfilters could be added as an extra mode to trigger's main().

MikeLing:
so, is that mean we don't need filter as a command, and we only use --include/--exclude to distinguish if we are using filter or not?

adusca:
@MikeLing That's right.

armenzg:
I see more code changes after @adusca's last comment.
Is this ready to be looked at?

MikeLing:
Not yep at that time(haven't test it)
But I think it's good to go now :) @adusca and @armenzg r?

mozci/scripts/trigger.py
```diff
@@ -286,38 +304,79 @@ def main():
...
+                                           'fromchange': revlist[-1],
+                                           'tochange': revlist[0],
+                                           'filter-searchStr': buildername}))
+
+    # Mode #4: retriggers N times every job that
+    # matches --includes and doesn't match --exclude .
+    else:
+        filter_in = options.includes.split(',') + [options.repo_name] if options.includes else []
+        filter_out = options.exclude.split(',') if options.exclude else []
+        buildernames = filter_buildernames(
```
adusca, line 340: Good job! But there is still some repetition going on here. Ideally, instead of having trigger_range in the if/else block we would only do:

```
if options.includes or options.exclude:
    filter_in = 
    options.buildernames = ...
```

and keep the rest of the code almost unchanged.

MikeLing:
@adusca ?
BTW, How could I add the "need review" tag?

MikeLing:
Oh, this PR need updated anyway after this (https://github.com/mozilla/mozilla_ci_tools/pull/409) been landed. But you @armenzg and @adusca could review it first and leave the reset of work to me ;)

armenzg:
Looking at this.

mozci/scripts/trigger.py
```diff
@@ -143,9 +160,11 @@ def validate_options(options):
     """
     error_message = ""
     if (not options.buildernames and
-       (not options.coalesced and not options.fill_revision and not options.trigger_tests_only)):
+       (not options.coalesced and not options.fill_revision and not options.trigger_tests_only) and
```
armenzg: I realized this is getting harder to read.
Could you please change it to `not (A OR B OR ...) ?

mozci/scripts/trigger.py
```diff
@@ -171,6 +190,10 @@ def validate_options(options):
         if options.fill_revision:
             error_message = "You should not pass --fill-revision " \
                             "when you use --trigger-only-test-jobs"
+    if options.exclude or options.includes:
+        if not options.repo_name:
+            error_message = "A branch name is mandatory with --exclude or --include. "\
```
armenzg: s/branch/repo_name/

mozci/scripts/trigger.py
```diff
@@ -285,19 +308,68 @@ def main():
...
+        filters_out = []
+
+        if options.exclude:
+            filters_out = options.exclude.split(',')
+
+        buildernames = filter_buildernames(
+            buildernames=query_builders(repo_name=options.repo_name),
+            include=filters_in,
+            exclude=filters_out
+        )
```
armenzg, line 344: Good :)

mozci/scripts/trigger.py
```diff
@@ -285,19 +308,68 @@ def main():
...
-            back_revisions=options.back_revisions,
-            delta=options.delta,
-            from_rev=options.from_rev,
-            backfill=options.backfill,
-            skips=options.skips,
-            max_revisions=options.max_revisions)
+    else:
+        buildernames = options.buildernames
+
+    if len(buildernames) == 0:
```
armenzg: Can you please move this block into https://github.com/mozilla/mozilla_ci_tools/pull/402/files#diff-14a3d986e72396dbc0d1a2774fee1046R313?

mozci/scripts/trigger.py
```diff
@@ -285,19 +308,68 @@ def main():
...
+    if cont.lower() == 'd':
+        LOG.info("The following jobs will be triggered: \n %s" % '\n'.join(buildernames))
+        cont = raw_input("Do you wish to continue? y/n ")
+
+    if cont.lower() != 'y':
+        exit(1)
+
+    for buildername in buildernames:
+        # We don't need determine a revision list for reigger range
+        # when we work under the mode 4.
```
armenzg: Please fix this sentence to this:
"We don't need to determine a revision list for trigger range when use includes/excludes."

mozci/scripts/trigger.py
```diff
@@ -285,19 +308,68 @@ def main():
...
+                rev=revision,
+                back_revisions=options.back_revisions,
+                delta=options.delta,
+                from_rev=options.from_rev,
+                backfill=options.backfill,
+                skips=options.skips,
+                max_revisions=options.max_revisions)
+
+            if revlist:
+                LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
```
armenzg: Do not add this inside of the for/loop or we will get one per builder.
Also get rid of the "filter-search".

mozci/scripts/trigger.py
```diff
@@ -285,19 +308,68 @@ def main():
...
+        LOG.info("The following jobs will be triggered: \n %s" % '\n'.join(buildernames))
+        cont = raw_input("Do you wish to continue? y/n ")
+
+    if cont.lower() != 'y':
+        exit(1)
+
+    for buildername in buildernames:
+        # We don't need determine a revision list for reigger range
+        # when we work under the mode 4.
+        if not (options.includes or options.exclude):
```
armenzg: Could you please put this if clause outside of the loop?

mozci/scripts/trigger.py
```diff
@@ -285,19 +308,68 @@ def main():
         )
         return
 
+    # Mode #4: retriggers N times every job that
```
armenzg: Could you invert this if/else clause so mode 4 shows up after mode 3?
Right now mode 4 is defined in line 313 while mode 3 starts in line 326.

mozci/scripts/trigger.py
armenzg: Ignore this comment.

mozci/scripts/trigger.py
```diff
@@ -133,6 +134,22 @@ def parse_args(argv=None):
                         dest="trigger_tests_only",
                         help="Schedule all missing tests for existing builds.")
 
+    parser.add_argument('--i', "--includes",
```
armenzg, line 138: Please add a note like in here:
https://github.com/MikeLing/mozilla_ci_tools/blob/issuefix-383/mozci/scripts/trigger.py#L115

mozci/scripts/trigger.py
```diff
@@ -133,6 +134,22 @@ def parse_args(argv=None):
...
+
+    parser.add_argument('--e', "--exclude",
+                        dest="exclude",
+                        type=str,
+                        help="comma-separated treeherder filters to exclude.")
+
+    parser.add_argument("--existing-only",
+                        action="store_false",
+                        dest="existing_only",
+                        help="Only trigger test jobs if the build jobs already exists.")
```
armenzg: s/jobs/job/ <- singular

armenzg:
Hi @MikeLing this is looking good.
There were some issues that were from before (typos and others) but this is shaping up well.

Could you also look into dropping this?
https://github.com/mozilla/mozilla_ci_tools/blob/master/setup.py#L10

/me adjusts labels

mozci/scripts/trigger.py
MikeLing: hmm, I don't think so. I know put this if clause inside the loop will cause a lot of unnecessary judge for **include** and **exclude**. But if we move this if clause outside of the loop, it will be:
<pre><code>if options.includes or options.exclude:
    revlist = [revision]
else:
    for buildername in buildernames:
            revlist = determine_revlist(
            ..............
**#because we still need buildname for trigger_range, so we need to loop buildername again.**
for buildername in buildernames:
      try:
            trigger_range(
                buildername=buildername,
                revisions=revlist,
                times=options.times,
                dry_run=options.dry_run,
                files=options.files,
                trigger_build_if_missing=options.trigger_build_if_missing
            )
        except Exception, e:
            LOG.exception(e)
            exit(1)
</code><pre>

Is that ok? or maybe we have a better way to do it?

mozci/scripts/trigger.py
MikeLing: hmm, sorry. I don't understand :( 
is that mean we could change the error message like "A branch/repo_name is mandatory with --exclude or --include." ?

armenzg: My apologies. I think this is something we do in IRC which is not obvious for everyone.

s/string1/string2/ comes from a syntax we use as part of vim or other editors.
It means the following:
 "Substitute string 1 for string 2"

mozci/scripts/trigger.py
armenzg: Let's keep two different loops. Mixing both logics makes reading the code much harder.

```
# Mode 3: Trigger by filters
if options.include or options.exclude:
    for buildername in ...
    # like in https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/triggerbyfilters.py#L131
else:
     # Mode 4: ...
    for buildername ...
    # like in https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/trigger.py#L290
```

MikeLing:
@armenzg ?

BTW, I try to not use git push -f because I know it will delete all comments, but it keeps telling me **Updates were rejected because the tip of your current branch is behind** even I "git pull --rebase" it. I apology if it makes you hard to review. I should talk to you on IRC before I push it, but I forget until I did it. Sorry :(

armenzg:
No worries! I appreciate that you try to avoid it.

I normally would do this:
- git checkout master
- git pull upstream master
- git checkout development_branch
- git rebase master

Is that what "git pull --rebase" would do?

I think what you can try next time is to "git merge master" and see if that fixes it.

mozci/scripts/trigger.py
```diff
@@ -288,38 +310,91 @@ def main():
...
-            skips=options.skips,
-            max_revisions=options.max_revisions)
-
-        try:
-            trigger_range(
+    if not (options.includes or options.exclude):
+        buildernames = options.buildernames
+
+    # Mode #4: retriggers N times every job that
+    # matches --includes and doesn't match --exclude .
```
armenzg: This instead:
Mode 4 - Schedule every builder matching --includes and does not match --exclude.

mozci/scripts/trigger.py
```diff
@@ -132,6 +133,22 @@ def parse_args(argv=None):
                         action="store_true",
                         dest="trigger_tests_only",
                         help="Schedule all missing tests for existing builds.")
+    # Mode #4: Retriggers N times every job that matches --includes and doesn't match --exclude.
```
armenzg: Use this instead:
Mode 4: Use --includes and --exclude flags to filter multiple buildernames.

mozci/scripts/trigger.py
```diff
@@ -288,38 +310,91 @@ def main():
...
+        if options.exclude:
+            filters_out = options.exclude.split(',')
+
+        buildernames = filter_buildernames(
+            buildernames=query_builders(repo_name=repo_name),
+            include=filters_in,
+            exclude=filters_out
+        )
+        if len(buildernames) == 0:
+            LOG.info("0 jobs match these filters. please try again.")
```
armenzg: There is a "return" in here:
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/triggerbyfilters.py#L113

Why is the return missing?

mozci/scripts/trigger.py
```diff
@@ -288,38 +310,91 @@ def main():
...
+    else:
+        cont = raw_input("%i jobs will be triggered, do you wish to continue? \
+                          y/n/d (d=show details) " % len(buildernames))
+
+    if cont.lower() == 'd':
+        LOG.info("The following jobs will be triggered: \n %s" % '\n'.join(buildernames))
+        cont = raw_input("Do you wish to continue? y/n ")
+
+    if cont.lower() != 'y':
+        exit(1)
```
armenzg: Can you please indent the block from line 333 to be part of the 'else' clause in line 318?

In fact, I belive from line 318 to here should be a call to a function something like "def determine_filtered_list_of_builders()".

mozci/scripts/trigger.py
```diff
@@ -288,38 +310,91 @@ def main():
...
+                    buildername=buildername,
+                    revisions=revlist,
+                    times=options.times,
+                    dry_run=options.dry_run,
+                    files=options.files,
+                    trigger_build_if_missing=options.trigger_build_if_missing
+                )
+            except Exception, e:
+                LOG.exception(e)
+                exit(1)
```
armenzg: I think this whole if/else block could be reduced to only one block.

You can run "determine_rev_list" even for the case when we only have one builder since it returns this:
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/trigger.py#L226

> revlist = [rev]

You can make the printing of the treeherder link to work with this:

```
LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
                         urllib.urlencode({'repo': repo_name,
                                           'fromchange': revlist[-1],
                                           'tochange': revlist[0],
                                           'filter-searchStr': buildername}))
```

It works for the cases of the revlist being of length one versus longer.

armenzg:
We're almost there. I'm sorry that this is being so much work to get it working nicely.
This script has become rather large and complex.

Thanks @MikeLing for your perseverance!

mozci/scripts/trigger.py
MikeLing: Hmm, but the Log info of trigger is different from the one of triggerbyfilter. we don't need the 'filter-searchStr' in trigger operation. Maybe we could use something like:

<pre><code>
    for buildername in buildernames:
        revlist = determine_revlist(
                repo_url=repo_url,
                ........
                max_revisions=options.max_revisions)

        if revlist:
            if options.includes or options.exclude:
                LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
                         urllib.urlencode({'repo': repo_name,
                                           'fromchange': revlist[-1],
                                           'tochange': revlist[0],
                                           'filter-searchStr': buildername}))
            else:
                LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
                         urllib.urlencode({'repo': repo_name,
                                           'revision': revision}))

        try:
            trigger_range(
                builde.......
            )
</code></pre>

mozci/scripts/trigger.py
MikeLing: > In fact, I belive from line 318 to here should be a call to a function something like "def determine_filtered_list_of_builders()".

hmm, I would just let it be there rather than def a function and call it if the determine_filtered_list_of_builders() will be only used in here.

mozci/scripts/trigger.py
armenzg: Good catch!
Perhaps a simple function like _print_treeherder_link() would make this block a little less crowded. Thanks!

mozci/scripts/trigger.py
armenzg: It is mainly because the main() function is getting very crowded and makes reading the code harder.

Do you agree with regards to the indentation issue?

MikeLing: yep, I had address them already(haven't push them yet) :)

mozci/scripts/trigger.py
```diff
@@ -132,6 +133,38 @@ def parse_args(argv=None):
...
+                        dest="exclude",
+                        type=str,
+                        help="comma-separated treeherder filters to exclude.")
+
+    parser.add_argument("--existing-only",
+                        action="store_false",
+                        dest="existing_only",
+                        help="Only trigger test job if the build jobs already exists.")
+
+    parser.add_argument('--i', "--includes",
```
armenzg, line 138: I believe the next 14 lines are duplicated code from the previous 15 lines.

mozci/scripts/trigger.py
```diff
@@ -313,13 +427,22 @@ def main():
...
             exit(1)
 
-        if revlist:
-            LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
-                     urllib.urlencode({'repo': repo_name,
-                                       'fromchange': revlist[-1],
-                                       'tochange': revlist[0],
-                                       'filter-searchStr': buildername}))
-
+        LOG.info('https://treeherder.mozilla.org/#/jobs?%s' %
```
armenzg: Do we need the next 15 lines? Otherwise, we're iterating the buildernames' list twice.

armenzg:
I don't know if the PR was ready to be looked at or not since I didn't see a comment saying so, however, I decided to refresh my memory by looking what had changed since last time.

Please let me know when you make more changes for me to look at.

mozci/scripts/trigger.py
MikeLing: No, absolutely not :)
Sorry, it must be a mistake when I try to fix the conflict caused by pull and merge. I'm not familiar enough to do push without squeeze.

PR #403, opened 2015-12-04
allow_using_existing_buildbot_build

armenzg:


armenzg:
r? @adusca 

It won't work for now without Mozharness changes but I would like to get the code at least in.

adusca:
r+

PR #407, opened 2015-12-15
Add missing None

adusca:
This fixes the error message we are getting from pulse_actions. r? @armenzg


adusca:
This should fix bug 1282388. I also noticed some KeyError exceptions, and it turns out we were never using properties['testsUrl'], so I changed that.

armenzg:
Very subtle! Thanks Alice!

PR #411, opened 2015-12-18
Setup logging immediately after parsing args

upsuper:
The function `query_repo_url` also produces some log which shouldn't be missed. Logging should be set up immediately after parsing args, otherwise there is a chance we miss something.


armenzg:
Thank you very much!

Issue #410, opened 2015-12-18
"That repository does not exist." is reported if --repo-name is not specified explicitly


upsuper:
It seems to me a while before, mozci was able to deduce the repo name itself without specifying anything. But after I upgraded to the latest version today, when I tried to run `mozci-trigger -b "Ubuntu VM 12.04 try opt test mochitest-e10s-browser-chrome-2" -r 18ecc556cb92`, it reports
```text
Traceback (most recent call last):
  File "/usr/local/bin/mozci-trigger", line 9, in <module>
    load_entry_point('mozci==0.23.2', 'console_scripts', 'mozci-trigger')()
  File "/usr/local/lib/python2.7/site-packages/mozci/scripts/trigger.py", line 237, in main
    repo_url = query_repo_url(options.repo_name)
  File "/usr/local/lib/python2.7/site-packages/mozci/repositories.py", line 41, in query_repo_url
    return query_repository(repo_name)["repo"]
  File "/usr/local/lib/python2.7/site-packages/mozci/repositories.py", line 34, in query_repository
    raise MozciError("That repository does not exist.")
mozci.errors.MozciError: That repository does not exist.
```

It only worked after I appended `--repo-name try` to the argument list.

This behavior doesn't make sense to me, because the buildername already includes the repo name, and I shouldn't be required to repeat that.

armenzg:
You're right. Thanks for reporting this.

@MikeLing do you have time to look into helping Xidorn?

adusca:
https://github.com/mozilla/mozilla_ci_tools/pull/390#issuecomment-162002475

MikeLing:
yep, I will look more into this :)

armenzg:
@upsuper we've fixed this in the dot release I just made.
Thanks for reporting!

PR #414, opened 2015-12-20
Issue 410 - Do not fail when --repo-name does not need to be set

MikeLing:


MikeLing:
This PR is about fix https://github.com/mozilla/mozilla_ci_tools/issues/410 
The miss repo name error is about somehow it require **repo name** to generate **repo_url** before we actually query it from buildname.

MikeLing:
@armenzg and @adusca r?

armenzg:
Looking at this.

mozci/scripts/trigger.py
```diff
@@ -254,6 +253,7 @@ def main():
     if not options.repo_name:
         options.repo_name = query_repo_name_from_buildername(options.buildernames[0])
 
+    repo_url = query_repo_url(options.repo_name)
```
armenzg: Notice what the developer filed.
He said that he did not set --repo-name; with your code change, what would happen if he does the same thing?

MikeLing: Hmm, I think the developer said he don't think mozci should require repo_name for trigger because it should deduce it from buildname. And I had tested this by 

> mozci-trigger -b "Ubuntu VM 12.04 try opt test mochitest-e10s-browser-chrome-2" -r 18ecc556cb92
> which is @upsuper mentioned in that issue, and it works for me now.

armenzg: You're very right. Let me comment what I think should be fixed as well.

mozci/scripts/trigger.py
```diff
@@ -254,6 +253,7 @@ def main():
     if not options.repo_name:
         options.repo_name = query_repo_name_from_buildername(options.buildernames[0])
```
armenzg: This is bad habit (setting values that the user did not pass) and I should have not allowed this to be commited.

Can you please add this?
'''
if not options.repo_name:
     if not options.repo_name:
         repo_name = query_repo_name_from_buildername(options.buildernames[0])
     else:
         repo_name = options.repo_name

repo_url = query_repo_url(repo_name)
'''
Also make any changes to substitute options.repo_name for repo_name

Thanks!

armenzg:
Feel free to squash in your next push since I think we will only need one more commit in here.

Thanks @MikeLing for pointing out my mistake!

mozci/scripts/trigger.py
MikeLing: Hmm,I guess this two 

> if not options.repo_name:
> if not options.repo_name:

is a mistype or something? ;)

MikeLing:
@armenzg r?

PR #417, opened 2016-01-07
Add SETA awarness to mozci

### LINKED ISSUE #366: Give SETA awareness to mozci

F3real:
Fixes #366

### DIFF
--- a/test/test_platforms.py
+++ b/test/test_platforms.py
@@ -15,6 +15,8 @@ from mozci.platforms import (
     get_associated_platform_name,
     get_buildername_metadata,
     get_downstream_jobs,
+    get_SETA_info,
+    get_SETA_interval_dict,
     filter_buildernames,
     find_buildernames,
     is_downstream,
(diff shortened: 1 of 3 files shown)


test/test_platforms.py
```diff
@@ -119,6 +121,34 @@ def test_include_exclude(self):
...
+    def test_parse_correctly(self, fetch_allthethings_data):
+        """get_SETA_interval_dict should return a dict with correct SETA intervals."""
+        fetch_allthethings_data.return_value = MOCK_ALLTHETHINGS
+        self.assertEquals(
+            get_SETA_interval_dict(),
+            {"Rev4 MacOSX Snow Leopard 10.6 fx-team debug test cppunit": [7, 3600]}
+        )
+
+
+class TestGetSETAInfo(unittest.TestCase):
```
armenzg: The only change I would do is to only have 1 class but two tests. You can call the class TestSETA.

armenzg:
Hi @F3real this is great!
I would only do that one change I mentioned.
My apologies for having taken so long to review this. I hope that having dealt with other review faster could somewhat redeem me! sorry again.

armenzg:
Let me know when you fix that class and I will merge this.

F3real:
I've made changes. @armenzg

armenzg:
Wonderful! Thanks!

PR #419, opened 2016-01-12
Update treeherder-client dependency from >=1.4 to >=2.0.1

edmorley:
To ensure consumers of mozci don't use deprecated versions of TreeherderClient if they are re-using virtualenvs.

Notably 2.0.1 includes an API URL fix that will prevent 404s once non-canonical URLs are disabled in bug 1234233.


armenzg:
Thank you Ed!

PR #431, opened 2016-01-29
Re-factor _status_summary() to be less hard-coded

MikeLing:
Hi @armenzg , I haven't test this PR yet. Just want to make sure it's on the right direction :)

Thank u


armenzg:
Looking good! You need to go through the tox failures and that should be it.

MikeLing:
@armenzg r?

PR #433, opened 2016-02-10
re-trigger all orange jobs for a particular revision

MikeLing:
Here is what I got by: (mozicitool_env)MikeLings-MacBook-Pro-2:mozilla_ci_tools mikeling$ python mozci/scripts/trigger.py --revision a1e682c00ff3 --repo-name try --failed-jobs

> 02/10/2016 11:08:54 root INFO:  Setting INFO level
> [u'Linux x86-64 try leak test build', u'Ubuntu VM 12.04 x64 try debug test mochitest-2', u'Ubuntu VM 12.04 x64 try debug test mochitest-4']
> 02/10/2016 11:09:02 root INFO:  https://treeherder.mozilla.org/#/jobs?repo=try&revision=a1e682c00ff34c054a1217a8f057038cc8f8ffbb
> 02/10/2016 11:09:02 mozci INFO: We want to have 1 job(s) of Linux x86-64 try leak test build on the following revisions: 
> 02/10/2016 11:09:02 mozci INFO:  - a1e682c00ff34c054a1217a8f057038cc8f8ffbb
> 02/10/2016 11:09:02 mozci INFO: 
> 02/10/2016 11:09:02 mozci INFO: === a1e682c00ff34c054a1217a8f057038cc8f8ffbb ===
> 02/10/2016 11:09:04 mozci INFO: We want to have 1 job(s) of Linux x86-64 try leak test build
> 02/10/2016 11:09:14 mozci INFO: We have 1 job(s) for 'Linux x86-64 try leak test build' which is enough for the 1 job(s) we want.
> 02/10/2016 11:09:14 root INFO:  https://treeherder.mozilla.org/#/jobs?repo=try&revision=a1e682c00ff34c054a1217a8f057038cc8f8ffbb
> 02/10/2016 11:09:14 mozci INFO: We want to have 1 job(s) of Ubuntu VM 12.04 x64 try debug test mochitest-2 on the following revisions: 
> 02/10/2016 11:09:14 mozci INFO:  - a1e682c00ff34c054a1217a8f057038cc8f8ffbb
> 02/10/2016 11:09:14 mozci INFO: 
> 02/10/2016 11:09:14 mozci INFO: === a1e682c00ff34c054a1217a8f057038cc8f8ffbb ===
> 02/10/2016 11:09:14 mozci INFO: We want to have 1 job(s) of Ubuntu VM 12.04 x64 try debug test mochitest-2
[...]

Issue #426, opened 2016-01-22
Make determine_trigger_job be able to work without BuildApi()


armenzg:
Currently, to determine what to trigger, we need to talk with BuildApi() and grab all sorts of buildjson files. Fetching these files over network end up taking too much time.

We should be able to change determine_trigger_job() to work with treeherder [1]. Even if treeherder data can be less accurate.

In my next quarter I hope to fix the accuracy of Treeherder by switching Buildbot jobs to be scheduled via BBB.

@MikeLing let me know if you would be interested on this bug.

[1] https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/mozci.py#L139

MikeLing:
Oh, sorry! I just see this issue and somehow I miss the notification from github :(

I haven't read it in detail, but I would like to happy of course ;)

armenzg:
Yay! Thanks @MikeLing . This might be a tough one (or an ugly issue :P).

armenzg:
Adding notes about using TaskCluster to find files:
########################################
<armenzg> mikeling, what I was telling you about TC, we can use the index to find tasks with artifacts
<armenzg> *every* buildbot job build uploads to TC
<armenzg> it's a neat dark secret
<armenzg> for instance, for a given revision
<armenzg> e.g. https://tools.taskcluster.net/index/#gecko.v2.mozilla-inbound.revision.0003682e05b93a1ce2f226665b848d25f12f3919.firefox/gecko.v2.mozilla-inbound.revision.0003682e05b93a1ce2f226665b848d25f12f3919.firefox.linux64-debug
<armenzg> and a platform
<armenzg> we can get to this task https://tools.taskcluster.net/task-inspector/#t1dd870CQsWqlVFB_qo0ag/
<armenzg> which has *all* the files we want https://tools.taskcluster.net/task-inspector/#t1dd870CQsWqlVFB_qo0ag/0
<armenzg> :)
<armenzg> in other words, given a Linux builder that comes from treeherder's data
<armenzg> we could determine the platform (e.g. linux64-debug)
<armenzg> we can add the functionality to say that it should be "firefox"
<armenzg> and we can establish the URL to the index
<armenzg> and the index can tell us which task contains all the artifacts

MikeLing:
I got some progress on this :)

[...]

PR #436, opened 2016-02-19
Fix for find_all_jobs_by_status()

MikeLing:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/436)

### DIFF
--- a/mozci/query_jobs.py
+++ b/mozci/query_jobs.py
@@ -309,7 +309,7 @@ class TreeherderApi(QueryApi):
                 continue
             if job_status == status:
                 if job['build_system_type'] == 'taskcluster':
-                    job_name = job['job_type_name'].split(' ')[-1]
+                    job_name = job['job_type_name']
                 else:
                     job_name = job['ref_data_name']
                 builder_names.append(job_name)
--- a/mozci/query_jobs.py
+++ b/mozci/query_jobs.py
@@ -309,8 +309,8 @@ class TreeherderApi(QueryApi):
                 continue
             if job_status == status:
                 if job['build_system_type'] == 'taskcluster':
-                    builder_name = job['job_type_name'].split(' ')[-1]
+                    job_name = job['job_type_name'].split(' ')[-1]
                 else:
-                    builder_name = job['ref_data_name']
-                builder_names.append(builder_name)
+                    job_name = job['ref_data_name']
+                builder_names.append(job_name)
         return builder_names
--- a/mozci/query_jobs.py
+++ b/mozci/query_jobs.py
@@ -308,5 +308,9 @@ class TreeherderApi(QueryApi):
             except TreeherderError:
                 continue
             if job_status == status:
-                builder_names.append(job['ref_data_name'])
+                if job['build_system_type'] == 'taskcluster':
+                    builder_name = job['job_type_name'].split(' ')[-1]
+                else:
+                    builder_name = job['ref_data_name']
+                builder_names.append(builder_name)
         return builder_names
--- a/mozci/scripts/trigger.py
+++ b/mozci/scripts/trigger.py
@@ -25,7 +25,7 @@ from mozhginfo.pushlog_client import (
 from mozci.utils.authentication import valid_credentials, get_credentials
 from mozci.utils.log_util import setup_logging
 from mozci.platforms import filter_buildernames
-from mozci.query_jobs import WARNING
+from mozci.query_jobs import WARNING, SUCCESS
 
 
 def parse_args(argv=None):
@@ -151,7 +151,7 @@ def parse_args(argv=None):
                         help="comma-separated treeherder filters to exclude.")
 
     parser.add_argument("--existing-only",
-                        action="store_false",
+                        action="store_true",
                         dest="existing_only",
                         help="Only trigger test job if the build jobs already exists.")
 
@@ -172,7 +172,7 @@ def validate_options(options):
     error_message = ""
     if not(options.buildernames or options.coalesced or options.fill_revision or
            options.trigger_tests_only or options.includes or options.exclude or
-           options.failed_jobs):
+           options.existing_only or options.failed_jobs):
         error_message = "A buildername is mandatory for all modes except --coalesced, " \
                         "--fill-revision, --trigger-only-test-jobs --include, --exclude" \
                         " and --failed-jobs. Use --buildername."
@@ -292,7 +292,6 @@ def main():
         LOG = setup_logging(logging.INFO)
 
     validate_options(options)
-
     if not valid_credentials():
         sys.exit(-1)
 
@@ -323,6 +322,9 @@ def main():
     else:
         mgr = BuildAPIManager()
 
+    trigger_build_if_missing = options.trigger_build_if_missing
+    if repo_name == 'try':
+        trigger_build_if_missing = False
     # Mode 1: Trigger coalesced jobs
     if options.coalesced:
         query_api = BuildApi()
@@ -370,6 +372,15 @@ def main():
             return
 
         if options.existing_only:
+            # we will query all successful job by given revision and only keep
+            # the those ones within the builder list been filted by exclude and include.
+            trigger_build_if_missing = False
+            existing_buildername = TreeherderApi().find_all_jobs_by_status(
+                repo_name=repo_name,
+                revision=revision,
+                status=SUCCESS)
+            buildernames = [buildername for buildername in existing_buildername
+                            if buildername in buildernames]
             cont = raw_input("The ones which have existing builds out of %i jobs will be triggered,\
                              do you wish to continue? y/n/d (d=show details) " % len(buildernames))
         else:
@@ -383,7 +394,7 @@ def main():
         if cont.lower() != 'y':
             exit(1)
 
-    # Mode 5: Use --existing-jobs or --failed-jobs to trigger jobs for particular revision
+    # Mode 5: Use --failed-jobs to trigger jobs for particular revision
     elif options.failed_jobs:
         buildernames = TreeherderApi().find_all_jobs_by_status(
             repo_name=repo_name,
@@ -419,7 +430,7 @@ def main():
                 times=options.times,
                 dry_run=options.dry_run,
                 files=options.files,
-                trigger_build_if_missing=options.trigger_build_if_missing
+                trigger_build_if_missing=trigger_build_if_missing
             )
         except Exception, e:
             LOG.exception(e)
--- a/mozci/scripts/trigger.py
+++ b/mozci/scripts/trigger.py
@@ -372,14 +372,15 @@ def main():
             return
 
         if options.existing_only:
-            # we will query all successful job by given revision and only keep
-            # the those ones within the builder list been filted by exclude and include.
+            # We query all succesful jobs for a given revision and filter
+            # them by include/exclude filters.
             trigger_build_if_missing = False
-            existing_buildername = TreeherderApi().find_all_jobs_by_status(
+            successful_jobs = TreeherderApi().find_all_jobs_by_status(
                 repo_name=repo_name,
                 revision=revision,
                 status=SUCCESS)
-            buildernames = [buildername for buildername in existing_buildername
+            # We will filter out all the existing job from those successful job we have.
+            buildernames = [buildername for buildername in successful_jobs
                             if buildername in buildernames]
             cont = raw_input("The ones which have existing builds out of %i jobs will be triggered,\
                              do you wish to continue? y/n/d (d=show details) " % len(buildernames))
--- a/mozci/scripts/trigger.py
+++ b/mozci/scripts/trigger.py
@@ -352,7 +352,7 @@ def main():
 
     # Mode #3: Trigger jobs based on revision list modifiers
     if not (options.includes or options.exclude or options.failed_jobs):
-        buildernames = options.buildernames
+        job_names = options.buildernames
 
     # Mode 4 - Schedule every builder matching --includes and does not match --exclude.
     elif options.includes or options.exclude:
@@ -362,12 +362,12 @@ def main():
         if options.exclude:
             filters_out = options.exclude.split(',')
 
-        buildernames = filter_buildernames(
+        job_names = filter_buildernames(
             buildernames=query_builders(repo_name=repo_name),
             include=filters_in,
             exclude=filters_out
         )
-        if len(buildernames) == 0:
+        if len(job_names) == 0:
             LOG.info("0 jobs match these filters. please try again.")
             return
 
@@ -380,16 +380,16 @@ def main():
                 revision=revision,
                 status=SUCCESS)
             # We will filter out all the existing job from those successful job we have.
-            buildernames = [buildername for buildername in successful_jobs
-                            if buildername in buildernames]
+            job_names = [buildername for buildername in successful_jobs
+                         if buildername in job_names]
             cont = raw_input("The ones which have existing builds out of %i jobs will be triggered,\
-                             do you wish to continue? y/n/d (d=show details) " % len(buildernames))
+                             do you wish to continue? y/n/d (d=show details) " % len(job_names))
         else:
             cont = raw_input("%i jobs will be triggered, do you wish to continue? \
-                              y/n/d (d=show details) " % len(buildernames))
+                              y/n/d (d=show details) " % len(job_names))
 
         if cont.lower() == 'd':
-            LOG.info("The following jobs will be triggered: \n %s" % '\n'.join(buildernames))
+            LOG.info("The following jobs will be triggered: \n %s" % '\n'.join(job_names))
             cont = raw_input("Do you wish to continue? y/n ")
 
         if cont.lower() != 'y':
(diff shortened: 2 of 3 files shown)


mozci/query_jobs.py
```diff
@@ -308,5 +308,11 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
             except TreeherderError:
                 continue
             if job_status == status:
-                builder_names.append(job['ref_data_name'])
+                if job['build_system_type'] == 'taskcluster':
+                    platform = '[TC]'
+                    job_type_name = job['job_type_name'].replace('-', '')
```
armenzg: Could you take "job_type_name" as is without any replacing or splitting?

We will have to re-evaluate this in the future.
I'm also hoping to standarize the names for TaskCluster tasks in Q2.

mozci/query_jobs.py
```diff
@@ -308,5 +308,11 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
...
+                    builder_name = job_type_name.split(platform)[-1]
+                else:
+                    builder_name = job['job_type_name']
+                builder_names.append(builder_name)
```
armenzg: I would take this opportunity to rename builder_names to job_uuid.

MikeLing:
I found this PR(https://github.com/mozilla/mozilla_ci_tools/pull/433) actually doesn't work. The main ideal in there is about to query 'ref_data_name' and use it as buildername(https://github.com/mozilla/mozilla_ci_tools/pull/433/files#diff-c98ed40e085d771e97c54278bf4a8d64R311).

But, I somehow forget the 'ref_data_name' of 'taskcluster' job is the job's signature :( Sorry about that.
And I believe this patch should fix it and make it works :)

Thank you.

mozci/query_jobs.py
MikeLing: got it

mozci/scripts/trigger.py
```diff
@@ -370,6 +369,12 @@ def main():
             return
 
         if options.existing_only:
+            exiting_buildernames = TreeherderApi().find_all_jobs_by_status(
```
armenzg: existing_jobs instead of existing_buildernames.

In general we're going to switch from using the term buildernams for 'jobs' (since we're adding TC to the mix).

mozci/query_jobs.py
MikeLing: > Could you take "job_type_name" as is without any replacing or splitting?

hmm, is that mean we could use buildername like '[TC] - ESLint'? I'm not sure about that

> We will have to re-evaluate this in the future.

yeah. that's really annoying.

armenzg: The way it will work in the future is that there will be a full TaskCluster graph with the original names.
I can use such dictionary/graph as a reference book and I can then re-submit tasks by making small changes but I have to keep "name" as-is.

MikeLing: hmm, I haven't got the point. sorry :(

I foully understand we will switch to TC in the further and we may not need buildername anymore at that point. But how about now? AFAIK, we can't use job's job_type_name to trigger a job if it's a taskcluster job. Maybe we could back to this issue after this [one](https://github.com/mozilla/mozilla_ci_tools/issues/426)  been first fixed?

armenzg: Sure. It works for me.

armenzg:
I think the only thing left is the typo for "exi**s**ting" instead "exiting"
After that please merge this.

MikeLing:
@armenzg haven't test it yet, so ask feedback for now :)

test/test_query_jobs.py
```diff
@@ -274,3 +353,39 @@ def test_matching_jobs_invalid(self):
...
+class TestTestTreeherderApiGetAllJobsByStatus(unittest.TestCase):
+    """Test TreeherderApi fetch all jobs(buildernames for now) by job's status."""
+
+    def setUp(self):
+        self.query_api = TreeherderApi()
+        self.repo_name = 'repo_mock'
+        self.revision = 'revision_mock'
+
+    @patch('mozci.query_jobs.TreeherderApi.get_all_jobs',
+           return_value=json.loads(MOCK_JOBS % {'result': "success", 'state': "completed"}))
```
armenzg, line 367: I like how you change the values on the fly.
Clever solution!

mozci/scripts/trigger.py
```diff
@@ -370,6 +369,12 @@ def main():
             return
 
         if options.existing_only:
+            existing_buildername = TreeherderApi().find_all_jobs_by_status(
+                repo_name=repo_name,
+                revision=revision,
+                status=SUCCESS)
+            buildernames = [buildername for buildername in existing_buildername
+                            if buildername in buildernames]
```
armenzg: I'm having trouble parsing exactly what we're trying to do in here. I think this script has become a little too complex and I haven't looked at it for a while.

What I believe you're trying to do is to select all jobs that have completed successfully and only keep the buildernames for those jobs that have completed SUCCESSfully.

However, if I recall correctly, when someone uses --existing-only is because they want to trigger a bunch of tests (specified through --include --exclude) and make sure that they don't trigger **build** job which are missing.

I think this used to be taken care of by trigger_build_if_missing:
https://github.com/mozilla/mozilla_ci_tools/commit/cebdb6615f6cec3c0bebd46aba470e1aa93c7a1a#diff-fd91cda9284f16c1f7364ee281591b09L134

I think we should add those 3 lines back in here. I'm still having trouble being sure of it.

I believe you're also trying to create an accurate list of what is actually going to be scheduled.
Perhaps we could look at the list of _build_ jobs which have completed SUCCESSfuly and remove them if they're not in such list.

MikeLing:
@armenzg Hey, somehow I can't query buildername [here](https://github.com/mozilla/mozilla_ci_tools/pull/436/files#diff-14a3d986e72396dbc0d1a2774fee1046R365)(when I use pdb test it, the flow comes to that line and stuck in there) I can't figure out why. Could you test this patch please? Maybe it's my local network problem(It happens sometimes in China :( )
Sorry

armenzg:
No worries. I will.

mozci/scripts/trigger.py
armenzg: I think there is a bug in here as I mentioned on IRC.

Remember that we only care about _build_ jobs.

Without --existing-only we get 55 builders [1], however, when we use --existing-only we should only get 7 builders (since the Linux opt and debug builds exist).

[1] https://gist.github.com/armenzg/78fc0438da1bb9f3522c
[2] https://gist.github.com/armenzg/17a912072c36abd38552

armenzg:
Hi @MikeLing do you want me to test this? I completely lost track of this. My apologies.

I wish GH would have a way to help me not forget these.

MikeLing:
@armenzg r?

> (mozicitool_env)localhost:mozilla_ci_tools mikeling$ python mozci/scripts/trigger.py --repo try --revision 76dffca6207d --dry-run --include "chrome-1" --existing-only
> 04/03/2016 11:27:40 root INFO:  Setting INFO level
> The ones which have existing builds out of 7 jobs will be triggered,                             do you wish to continue? y/n/d (d=show details) d
> 04/03/2016 11:27:58 root INFO:  The following jobs will be triggered: 
>  Ubuntu VM 12.04 x64 try opt test mochitest-devtools-chrome-1
> Ubuntu VM 12.04 x64 try opt test mochitest-e10s-devtools-chrome-1
> Ubuntu VM 12.04 x64 try opt test mochitest-e10s-browser-chrome-1
> Ubuntu VM 12.04 x64 try opt test mochitest-browser-chrome-1
> Ubuntu VM 12.04 x64 try debug test mochitest-browser-chrome-1
> Ubuntu VM 12.04 x64 try debug test mochitest-devtools-chrome-1
> Ubuntu VM 12.04 x64 try debug test mochitest-e10s-browser-chrome-1
> Do you wish to continue? y/n n

mozci/query_jobs.py
```diff
@@ -308,5 +308,9 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
             except TreeherderError:
                 continue
             if job_status == status:
-                builder_names.append(job['ref_data_name'])
+                if job['build_system_type'] == 'taskcluster':
+                    builder_name = job['job_type_name'].split(' ')[-1]
```
armenzg: Could you put an example of what the value of job_type_name can be for TaskCluster?

Could you please change builder_name for job_name? (a bit more generic)

Can you also add a docstring?

mozci/scripts/trigger.py
```diff
@@ -370,6 +372,15 @@ def main():
             return
 
         if options.existing_only:
+            # we will query all successful job by given revision and only keep
+            # the those ones within the builder list been filted by exclude and include.
```
armenzg: "We query all succesful jobs for a given revision and filter them by include/exclude filters."

mozci/scripts/trigger.py
```diff
@@ -370,6 +372,15 @@ def main():
...
+            trigger_build_if_missing = False
+            existing_buildername = TreeherderApi().find_all_jobs_by_status(
```
armenzg: Why the variable name "existing_buildername"?
Would "successful_jobs" make more sense? I'm asking since I'm having trouble comprehending it. Maybe a comment would help me?

mozci/scripts/trigger.py
```diff
@@ -370,6 +372,15 @@ def main():
...
+                repo_name=repo_name,
+                revision=revision,
+                status=SUCCESS)
+            buildernames = [buildername for buildername in existing_buildername
```
armenzg: I think adding a comment here would help (maybe a different variable name).

jobs? successful_jobs? requested_jobs? what do you think?

test/test_query_jobs.py
```diff
@@ -74,6 +74,85 @@
...
+    "signature": "3681e7bdec56673c80e85486dcfd66d4e2c57185",
+    "result_set_id": 27064,
+    "result": "%(result)s",
+    "machine_platform_os": "-",
+    "ref_data_name": "3681e7bdec56673c80e85486dcfd66d4e2c57185",
+    "machine_platform_architecture": "-",
+    "end_timestamp": 1455884086,
+    "build_platform": "lint",
+    "job_guid": "6657633e-a463-4955-bc9c-3a561b236308/0",
+    "job_type_name": "[TC] - ESLint",
```
armenzg: Could you please look for a TC job that does not contain only one word name? (Ignoring [TC])

I've found "[TC] Linux64 gtest"or "[TC] Linux64 web-platform-tests-e10s-6"

You can find jobs in here:
https://treeherder.mozilla.org/#/jobs?repo=mozilla-central&filter-searchStr=tc

test/test_query_jobs.py
```diff
@@ -274,3 +353,42 @@ def test_matching_jobs_invalid(self):
             self.query_api.get_matching_jobs(
                 "try", "146071751b1e",
                 'Invalid buildername'), [])
+
+
+class TestTestTreeherderApiGetAllJobsByStatus(unittest.TestCase):
```
armenzg: s/TestTest/Test/

armenzg:
This looks good.
Could you please don't squash your next push? I would like to see the small changes.
It will be on me to squash them rather than asking you to do it. Thanks!

mozci/query_jobs.py
MikeLing: > Could you put an example of what the value of job_type_name can be for TaskCluster?

Sure :) So for a job like:

> {u'submit_timestamp': 1456338218, u'build_system_type': u'taskcluster', u'machine_name': u'i-3f6ecefc', u'job_group_symbol': u'tc-M-e10s', u'job_group_name': u'Desktop mochitests', u'platform_option': u'debug', u'job_type_description': u'fill me', u'signature': u'2c6986f7d282789bafb62b6c673c52f3e281b87f', u'result_set_id': 91926, u'result': u'success', u'machine_platform_os': u'-', u'ref_data_name': u'2c6986f7d282789bafb62b6c673c52f3e281b87f', u'machine_platform_architecture': u'-', u'end_timestamp': 1456340350, u'build_platform': u'linux64', u'job_guid': u'564027ea-52a8-442f-9078-9cfa7f4ac2a4/0', u'job_type_name': u'[TC] Linux64 mochitest-browser-chrome-e10s-2', u'id': 17179381, u'platform': u'linux64', u'state': u'completed', u'job_type_id': 4494, u'build_os': u'-', u'option_collection_hash': u'32faaecac742100f7753f0c1d0aa0add01b4046b', u'who': u'mozilla-taskcluster-maintenance@mozilla.com', u'failure_classification_id': 1, u'job_type_symbol': u'bc2', u'reason': u'scheduled', u'job_group_description': u'fill me', u'tier': 2, u'job_coalesced_to_guid': None, u'running_eta': 2097, u'start_timestamp': 1456338220, u'build_architecture': u'-', u'last_modified': u'2016-02-24T18:59:12', u'build_platform_id': 106, u'job_group_id': 189}

The job data name will be: **job_type_name': u'[TC] Linux64 mochitest-browser-chrome-e10s-2**

so we can get buildname like: **mochitest-browser-chrome-e10s-2**

test/test_query_jobs.py
MikeLing: sure, we could just change it to "web-platform-tests-e10s-6" because it's mock data anyway :)

mozci/scripts/trigger.py
MikeLing: let's make it as successful_jobs :)

MikeLing: oh, I prefer to add a comment for it because we use buildernames in a lot of other places like [1]. I think we should keep consistency for it. 
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/scripts/trigger.py#L363

MikeLing:
@armenzg f? :)

mozci/query_jobs.py
```diff
@@ -309,8 +309,8 @@ def find_all_jobs_by_status(self, repo_name, revision, status):
                 continue
             if job_status == status:
                 if job['build_system_type'] == 'taskcluster':
-                    builder_name = job['job_type_name'].split(' ')[-1]
+                    job_name = job['job_type_name'].split(' ')[-1]
                 else:
-                    builder_name = job['ref_data_name']
-                builder_names.append(builder_name)
+                    job_name = job['ref_data_name']
+                builder_names.append(job_name)
```
armenzg, line 315: My bad. I should have also pointed out that I would have renamed builder_names as job_names to match.
Sorry!

test/test_query_jobs.py
```diff
@@ -370,8 +370,9 @@ def setUp(self):
     def test_successful_job(self, get_all_jobs, get_job_status):
         """Test TreeherderApi find_all_jobs_by_status with a successful job."""
         self.assertEqual(self.query_api.find_all_jobs_by_status(
-            self.repo_name, self.revision, SUCCESS), ["Ubuntu VM 12.04 x64 mozilla-inbound opt test mochitest-1",
-                                                      "ESLint"])
+            self.repo_name, self.revision, SUCCESS),
+            ["Ubuntu VM 12.04 x64 mozilla-inbound opt test mochitest-1",
+             "web-platform-tests-e10s-6"])
```
armenzg: I think we should keep the whole name rather than do string manipulations.
"[TC] - Linux64 web-platform-tests-e10s-6"
instead of
"web-platform-tests-e10s-6

armenzg:
Reviewed 2 of 3 files at r2.
Review status: 2 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

_[mozci/query_jobs.py, line 312 [r1]](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/436#-KEb8Xu7tmNScN0Xm51o-r1-312:-KEffan7LXTAqvmwB8WC:930521326) ([raw file](https://github.com/mozilla/mozilla_ci_tools/blob/18e0800c5e827cacb465b3fd103a4da10d1c8927/mozci/query_jobs.py#L312)):_
As I mentioned on the tests let's just take the whole job_type_name.

---

_[mozci/scripts/trigger.py, line 382 [r1]](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/436#-KEb8Xu87uDR1ji1NWMk-r1-382:-KEfeRDyqHtt9Zsx4uZg:-261118295) ([raw file](https://github.com/mozilla/mozilla_ci_tools/blob/18e0800c5e827cacb465b3fd103a4da10d1c8927/mozci/scripts/trigger.py#L382)):_
With regards to builders, buildernames and similar, I'm trying to move towards jobs and job_names since it is the middle ground between TaskCluster and Buildbot. Perhaps we should in one pass rather than little by little.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/436)_

armenzg:
I'm done giving feedback. Since you're close I'm happy to review it again later today or tomorrow morning. Thanks MikeLing!

---

Review status: 2 of 3 files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

_[test/test_query_jobs.py, line 133 [r1]](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/436#-KEb8Xu87uDR1ji1NWMl-r1-133:-KEfgAa4kc0xAfZ1LFJB:14247877) ([raw file](https://github.com/mozilla/mozilla_ci_tools/blob/18e0800c5e827cacb465b3fd103a4da10d1c8927/test/test_query_jobs.py#L133)):_
As I mentioned above let's take the whole name.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/436#-:-KEffyU1cXxRTrbkLfqf:534239080)_

MikeLing:
@armenzg r?

armenzg:
:lgtm:

---

Reviewed 3 of 3 files at r3.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/436#-:-KEgQ8XHw4-xEBhyDeTe:1417088721)_

PR #438, opened 2016-02-22
Get repositories data using treeherder

### LINKED ISSUE #416: Use treeherder to query information about repositories
We currently use buildapi to determine this information:
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/repositories.py
Unfortunately it requires credentials.

Treeherder has an API we should use instead:
https://treeherder.mozilla.org/api/repository
We should look if the treeherder client already has the code we can call to use this.

This would allow contributors without credentials to use mozci in dry-run mode (since querying about repositories requires credentials).

F3real:
Fixes #416

### DIFF
--- a/mozci/repositories.py
+++ b/mozci/repositories.py
@@ -7,15 +7,15 @@ import json
 import logging
 import os
 
-from buildapi_client import BuildapiAuthError, make_query_repositories_request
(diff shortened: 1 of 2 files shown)


armenzg:
Excellent! No issues found.

Issue #444, opened 2016-03-29
Indicate the buildername when submitting a BBB task instead of 'pulse_actions_graph'


armenzg:
In this graph:
https://tools.taskcluster.net/task-graph-inspector/#DdvObl_NR7KbNo48IUAI8w
You will see that every task is named 'pulse_graph'.
Instead, we should change this so it shows the buildername.

See one of the defined tasks:
https://queue.taskcluster.net/v1/task/flHe5_JPQK29jo6-h9H0Bw

Simply overwriting metadata['name'] should be sufficient.
https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/sources/buildbot_bridge.py#L98

We will need a new release for this.

@martiansideofthemoon this is an easy issue if you would like to get your feet wet fast.

martiansideofthemoon:
@armenzg ,
sorry, i was afk.. Yup i got a basic idea.. I've never worked with python unittests in the past. I've written unittests for Gaia though in JS. I think I understood some basic things. But I'll be honest, I didn't quite get the @patch part well. I think I wasn't patient enough with the documentation.

martiansideofthemoon:
@armenzg, yeah it makes sense now :smile: How was it relevant to the unit test i wrote?

armenzg:
It wasn't! It was my bad.

In general, we try to use @patch to prevent calls to allthethings.json since it is a bad file to write tests agains (hitting diff maxs).

armenzg:
I think this is now completed. Thanks @martiansideofthemoon

PR #445, opened 2016-03-29
Indicate the buildername when submitting a BBB task instead of 'pulse…

martiansideofthemoon:
…_actions_graph'
It locally passes `tox`
ref #444

### DIFF
--- a/mozci/sources/buildbot_bridge.py
+++ b/mozci/sources/buildbot_bridge.py
@@ -100,6 +100,10 @@ def _create_task(buildername, repo_name, revision, metadata=None, task_graph_id=
                           revision=revision,
                           name=buildername)
 
+    # The task's name is used in the task-graph-inspector to list all tasks
+    # and using the buildername makes it easy for a person to recognize each job.
+    metadata['name'] = buildername
+
     # XXX: We should validate that the parent task is a valid parent platform
     #      e.g. do not schedule Windows tests against Linux builds
     task = create_task(
--- a/test/test_buildbot_bridge.py
+++ b/test/test_buildbot_bridge.py
@@ -1,0 +1,28 @@
+"""This file contains tests for mozci/sources/buildbot_bridge.py."""
+import unittest
+from mozci.sources import buildbot_bridge
+
+
+class TestBuildbotBridge(unittest.TestCase):
+    """Test that buildbot bridge is correctly scheduling tasks"""
+    def setUp(self):
+        self.buildernames = ['Android armv7 API 15+ try debug build']
+        self.revision = '1ab622ac1706a0f5dfaf7734a1c56aa9d3502eec'
+        self.repo_name = 'try'
+        self.builders_graph, _ = buildbot_bridge.buildbot_graph_builder(
+            builders=self.buildernames,
+            revision=self.revision,
+            complete=False
+        )
+
+    def test_task_name_metadata_is_buildername(self):
+        self.graph = buildbot_bridge.generate_builders_tc_graph(
+            self.repo_name,
+            self.revision,
+            self.builders_graph
+        )
+        flag = False
+        for task in self.graph['tasks']:
+            if task['task']['metadata']['name'] == self.buildernames[0]:
+                flag = True
+        self.assertEquals(flag, True)


armenzg:
To write a test for it, you might need to write a decently long test with lots of prep [1](I had the code around from yesterday).

The code below could be made simpler by only calling generate_builders_tc_graph() in the test and check the name of the task there [4].

Once we have the returning graph we can assert that one of the tasks matches 'Android armv7 API 15+ try debug build' (graph['task'][<number>]['metadata']['name']
Or we can assert that the name in metadata for every task is a valid_builder()

Thanks for asking on how to test it!

[1]
Based on https://github.com/mozilla/pulse_actions/blob/master/pulse_actions/handlers/treeherder_runnable.py#L88-L107

```
buildernames=['Android armv7 API 15+ try debug build']
requester = 'email@me.com'
treeherder_link = 'https://treeherder.mozilla.org/#/jobs?repo=try&revision=1ab622ac1706'
dry_run=False
revision='1ab622ac1706a0f5dfaf7734a1c56aa9d3502eec'
repo_name='try'
builders_graph, _ = buildbot_bridge.buildbot_graph_builder(
            builders=buildernames,
            revision=revision,
            complete=False
)
mgr = TaskClusterBuildbotManager()
mgr.schedule_graph(
    repo_name=repo_name,
    revision=revision,
    metadata={
        'name': 'pulse_actions_graph',
        'description': 'Adding new jobs to push via pulse_actions/treeherder for %s.' % requester,
        'owner': requester,
        'source': treeherder_link,
    },
    builders_graph=builders_graph,
    dry_run=dry_run
)
```

[2] https://github.com/mozilla/mozilla_ci_tools/blob/master/mozci/ci_manager.py#L191-L196

martiansideofthemoon:
@armenzg , so is the fix correct? should I add a test in this PR itself? Also, not too sure where I need to write the test

armenzg:
The fix is correct. The test would go under the test/ directory.
test_buildbot_bridge.py (not existant).

armenzg:
Please add the test on this PR. That way you can get familiar with the unit testing in place.

martiansideofthemoon:
@armenzg, One link seems to be broken
[https://github.com/mozilla/mozilla_ci_tools/pull/I%20had%20the%20code%20around%20from%20yesterday](https://github.com/mozilla/mozilla_ci_tools/pull/I%20had%20the%20code%20around%20from%20yesterday)

test/test_buildbot_bridge.py
```diff
@@ -0,0 +1,30 @@
...
+builders_graph, _ = buildbot_bridge.buildbot_graph_builder(
+    builders=buildernames,
+    revision=revision,
+    complete=False
+)
+
+
+class TestBuildbotBridge(unittest.TestCase):
+    """Test that buildbot bridge is correctly scheduling tasks"""
+    def test_correct_name(self):
```
armenzg: Look at other tests to see what we do to patch allthethings.json

armenzg:
@martiansideofthemoon I looked a bit more closely and you should not see the failures you're getting in here:
https://travis-ci.org/mozilla/mozilla_ci_tools/builds/119314936

If you look at the tests, both are patching the data they should be receiving. I don't know what is going on.
Please poke me tomorrow if you're still hitting the issue.

Do you hit the issue without your test changes? (e.g. the master branch)

martiansideofthemoon:
@armenzg, I get all green on `master`
What exactly does `@patch` do? Unable to understand that

test/test_buildbot_bridge.py
```diff
@@ -0,0 +1,33 @@
...
+import unittest
+from mozci.sources import buildbot_bridge
+from mock import patch
+
+buildernames = ['Android armv7 API 15+ try debug build']
+requester = 'email@me.com'
+treeherder_link = 'https://treeherder.mozilla.org/#/jobs?repo=try&revision=1ab622ac1706'
+dry_run = False
+revision = '1ab622ac1706a0f5dfaf7734a1c56aa9d3502eec'
+repo_name = 'try'
```
armenzg: Please remove the variables that are not used below in the test.

armenzg:
I was surprised we would need @patch but I did not look further into it.

@patch allows you to modify the return value of a function. By doing this, you can control exactly what is going to return. It is important for cases where the return value is too big and makes it hard to control test failures. Does this help?

Here are the official docs:
https://docs.python.org/3.3/library/unittest.mock.html#unittest.mock.patch

martiansideofthemoon:
@armenzg , would it be necessary to use @patch here? The test worked correctly without @patch, but other tests were failing

mozci/sources/buildbot_bridge.py
```diff
@@ -99,7 +99,7 @@ def _create_task(buildername, repo_name, revision, metadata=None, task_graph_id=
         generate_metadata(repo_name=repo_name,
                           revision=revision,
                           name=buildername)
-
+    metadata['name'] = buildername
```
armenzg, line 105: Could you please add a note here? Say that the task's name is used in the task-graph-inspector to list all tasks and using the buildername makes it easy for a person to recognize each job.

armenzg:
I will test your code later when I'm not tethering.

Could you please address my comments and also let me know what happens when you run tox on your master branch? (I want to see if you still hit the same issue).

test/test_buildbot_bridge.py
```diff
@@ -0,0 +1,33 @@
...
+builders_graph, _ = buildbot_bridge.buildbot_graph_builder(
+    builders=buildernames,
+    revision=revision,
+    complete=False
+)
```
armenzg: Could you please move this code into a "setUp()" function inside of the class?

https://docs.python.org/2/library/unittest.html#unittest.TestCase.setUp

armenzg:
The code looks good.
Could you please squash the two commits into one before I merge it? [1]
I will merge it after that. Thanks @martiansideofthemoon !

[1] http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

armenzg:
I'm going to try to debug why we're having the Travis issue.

test/test_buildbot_bridge.py
```diff
@@ -0,0 +1,28 @@
...
+        self.buildernames = ['Android armv7 API 15+ try debug build']
+        self.revision = '1ab622ac1706a0f5dfaf7734a1c56aa9d3502eec'
+        self.repo_name = 'try'
+        self.builders_graph, _ = buildbot_bridge.buildbot_graph_builder(
+            builders=self.buildernames,
+            revision=self.revision,
+            complete=False
+        )
+
+    def test_correct_name(self):
```
armenzg: Can you please call test_task_name_metadata_is_buildername()?

armenzg:
I've filed an issue to improve the reliability of the tests, however, I don't think that's what you're hitting in Travis. I think there's something random happening in the pushlog_client.
https://github.com/mozilla/mozilla_ci_tools/issues/446

PR #447, opened 2016-03-31
Make determine_trigger_job be able to work without BuildApi()[DO NOT MERGE]

MikeLing:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/447)

### DIFF
--- a/mozci/mozci.py
+++ b/mozci/mozci.py
@@ -18,7 +18,7 @@ from mozci.platforms import (
     is_downstream,
     list_builders,
 )
-from mozci.sources import buildjson
+from mozci.sources import buildjson, tc
 from mozci.query_jobs import (
     PENDING,
     RUNNING,
@@ -49,6 +49,12 @@ QUERY_SOURCE = BuildApi()
 
 # Set this value to False in your tool to prevent any sort of validation
 VALIDATE = True
+
+INSTALLER_SUFFIXES = ('.apk',  # Android
+                      '.tar.bz2', '.tar.gz',  # Linux
+                      '.dmg',  # Mac
+                      '.installer-stub.exe', '.installer.exe', '.exe', '.zip',  # Windows
+                      )
 
 
 def disable_validations():
@@ -159,7 +165,7 @@ def determine_trigger_objective(revision, buildername, trigger_build_if_missing=
         return build_buildername, None, None
 
     # Let's figure out which jobs are associated to such revision
-    query_api = BuildApi()
+    query_api = TreeherderApi()
     # Let's only look at jobs that match such build_buildername
     build_jobs = query_api.get_matching_jobs(repo_name, revision, build_buildername)
 
@@ -192,7 +198,10 @@ def determine_trigger_objective(revision, buildername, trigger_build_if_missing=
             continue
 
         # Successful or failed jobs may have the files we need
-        files = _find_files(job)
+        taskId = tc.get_taskid(revision=revision,
+                               repo_name=repo_name,
+                               platform=job['build_platform'])
+        files = _find_files(taskId)
 
         if files != [] and _all_urls_reachable(files.values()):
             working_job = job
@@ -259,48 +268,40 @@ def determine_trigger_objective(revision, buildername, trigger_build_if_missing=
         return builder_to_trigger, None, None
 
 
-def _status_info(job_schedule_info):
-    # Let's grab the last job
-    complete_at = job_schedule_info["requests"][0]["complete_at"]
-    request_id = job_schedule_info["requests"][0]["request_id"]
-
-    # NOTE: This call can take a bit of time
-    return buildjson.query_job_data(complete_at, request_id)
-
-
-def _find_files(job_schedule_info):
+def _find_files(taskId):
     """
     Find the files needed to trigger a job.
 
     Raises MozciError if the job status doesn't have a properties key.
     """
     files = {}
+    test_url = None
+    test_packages_url = None
+    installer_url = None
 
-    job_status = _status_info(job_schedule_info)
-
-    if job_status is None:
-        LOG.warning("We should not have received an empty status")
+    job_artifacts = tc.get_artifacts(taskId)
+    if job_artifacts is None:
+        LOG.warning("We should not have received an empty job")
         return files
+    for f in job_artifacts['artifacts']:
+        if f['name'].endswith('tests.zip'):
+            if not test_url:
+                # str() because of unicode issues on mac
+                test_url = str(f['name'])
+                LOG.info("Found test url %s." % test_url)
 
-    properties = job_status.get("properties")
+        elif f['name'].endswith('test_packages.json'):
+            test_packages_url = str(f['name'])
+            LOG.info("Found a test packages url %s." % test_packages_url)
 
-    if not properties:
-        LOG.error(str(job_status))
-        raise MozciError("The status of the job is expected to have a "
-                         "properties key, however, it is missing.")
+        elif not any(f['name'].endswith(s) for s in ('code-coverage-gcno.zip',
+                                                     'crashreporter-symbols.zip')):
+            if any(f['name'].endswith(s) for s in INSTALLER_SUFFIXES):
+                installer_url = str(f['name'])
+                LOG.info("Found installer url %s." % installer_url)
 
-    LOG.debug("We want to find the files needed to trigger %s" %
-              properties["buildername"])
-
-    # We need the packageUrl, and one of testsUrl and testPackagesUrl,
-    # preferring testPackagesUrl.
-    if 'packageUrl' in properties:
-        files['packageUrl'] = properties['packageUrl']
-
-    if 'testPackagesUrl' in properties:
-        files['testsUrl'] = properties['testPackagesUrl']
-    elif 'testsUrl' in properties:
-        files['testsUrl'] = properties['testsUrl']
+    files['packageUrl'] = installer_url
+    files['testsUrl'] = test_packages_url if test_packages_url else test_url
 
     return files
 
--- a/mozci/sources/tc.py
+++ b/mozci/sources/tc.py
@@ -20,6 +20,7 @@ LOG = logging.getLogger('mozci')
 TC_TOOLS_HOST = 'https://tools.taskcluster.net'
 TC_TASK_INSPECTOR = "%s/task-inspector/#" % TC_TOOLS_HOST
 TC_TASK_GRAPH_INSPECTOR = "%s/task-graph-inspector/#" % TC_TOOLS_HOST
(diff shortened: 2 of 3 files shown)


MikeLing:
I haven't change any code in test_repositories.py, but why this error happened? https://travis-ci.org/mozilla/mozilla_ci_tools/builds/119732441#L218?

armenzg:
I'm investigating an issue with pushlog_client. I believe Mozilla's pushlog server started failing in the last two days. It might have become intermittently available. Ignore the Travis issue.

armenzg:
Is this ready to be reviewed? (since I don't see r? mentioned in the comment).

armenzg:
Just to be clear, we're allowing determine_trigger_objective() to work for BuildApi and TaskCluster (instead of dropping BuildAPI support). We might need two different functions, one for BuildApi and another one for TaskCluster.

MikeLing:
> Just to be clear, we're allowing determine_trigger_objective() to work for BuildApi and TaskCluster (instead of dropping BuildAPI support). We might need two different functions, one for BuildApi and another one for TaskCluster.

hmm, but buildbot job can been found in taskCluster right? So why do we still need two functions? And for the BuildApi, we could queue job from treeherder api instead of buildApi and so we can avoid query a bunch of json file from remote.

armenzg:
I am weiry of chaging core functionalities of mozci since it could affect pulse_actions. I prefer to transition little by little and keep backward compatibility.
However, if we have a fair amount of tests I would be less worried.

MikeLing:
Hi @armenzg , I'm still pretty confuse with the last comment. Sorry :(

I think, the main ideal in this issue, it's not about remove the BuildApi in the determine_trigger_job function, though we do use treeherderApi rather than use BuildApi now(https://github.com/mozilla/mozilla_ci_tools/pull/447/files#diff-fceb32314e6c82e12ca12ab409d93d8aL162).
The main goal we want to achieve in here, i think, is stop using json file and query those job data(package url and test url) from TaskCluster which will save us a lot of time. I haven't read some more code about how we achieve json file from remote. But i think we need to get ride of them anyway :)

mozci/mozci.py
```diff
@@ -159,7 +165,7 @@ def determine_trigger_objective(revision, buildername, trigger_build_if_missing=
         return build_buildername, None, None
 
     # Let's figure out which jobs are associated to such revision
-    query_api = BuildApi()
+    query_api = TreeherderApi()
```
armenzg, line 168: I think we can do set_query_source("treeherder") and use QUERY_SOURCE instead of query_api.

If we choose this route (only supporting one query approach), we will have to do a lot of testing of various scenarios. Either manual testing before releasing or adding tests calling determine_trigger_objective().

mozci/sources/tc.py
```diff
@@ -20,6 +20,7 @@
 TC_TOOLS_HOST = 'https://tools.taskcluster.net'
 TC_TASK_INSPECTOR = "%s/task-inspector/#" % TC_TOOLS_HOST
 TC_TASK_GRAPH_INSPECTOR = "%s/task-graph-inspector/#" % TC_TOOLS_HOST
+TC_NAMESPACE = "gecko.v2.%(reponame)s.revision.%(revision)s.firefox.%(platform)s"
```
armenzg, line 23: This will not work for mobile or b2g jobs.
e.g https://tools.taskcluster.net/index/#gecko.v2.mozilla-central.revision.538d248fa252a4100082fd9bc3fdc08d322cda22.b2g/gecko.v2.mozilla-central.revision.538d248fa252a4100082fd9bc3fdc08d322cda22.b2g

mozci/sources/tc.py
```diff
@@ -289,3 +290,19 @@ def generate_task_graph(scopes, tasks, metadata):
         'metadata': metadata
     }
     return task_graph
+
+
+def get_taskid(revision, repo_name, platform):
```
armenzg, line 295: Could you please add a docstring?

Could you also put three links? One for each product (firefox, b2g, mobile)
e.g. https://tools.taskcluster.net/index/#gecko.v2.mozilla-central.revision.538d248fa252a4100082fd9bc3fdc08d322cda22.b2g/gecko.v2.mozilla-central.revision.538d248fa252a4100082fd9bc3fdc08d322cda22.b2g

mozci/sources/tc.py
armenzg, line 23: We're going to have to write a test per platform (e.g. emulator-jb-opt, emulator-kk-opt) and product (e.g. b2g).

mozci/sources/tc.py
```diff
@@ -289,3 +290,19 @@ def generate_task_graph(scopes, tasks, metadata):
...
+    index = taskcluster_client.Index()
+    namespace = TC_NAMESPACE % {'revision': revision,
+                                'repo_name': repo_name,
+                                'platform': platform}
+    LOG.info("The namespace of this task is %s" % namespace)
```
armenzg, line 300: Could you please make this a debug message? No need for the average user to see it.

mozci/mozci.py
```diff
@@ -192,7 +198,10 @@ def determine_trigger_objective(revision, buildername, trigger_build_if_missing=
             continue
 
         # Successful or failed jobs may have the files we need
-        files = _find_files(job)
+        taskId = tc.get_taskid(revision=revision,
+                               repo_name=repo_name,
+                               platform=job['build_platform'])
```
armenzg, line 203: If the object job has revision and repo_name in its attributes, we could simply call _find_files(job) and internally determine the taskId; what do you think?

Does this new code still work for Buildbot jobs?
Did you test this code for Buildbot jobs on Treeherder? or against TaskCluster jobs running on Treeheder?

mozci/mozci.py
```diff
@@ -259,48 +268,40 @@ def determine_trigger_objective(revision, buildername, trigger_build_if_missing=
...
-    # Let's grab the last job
-    complete_at = job_schedule_info["requests"][0]["complete_at"]
-    request_id = job_schedule_info["requests"][0]["request_id"]
-
-    # NOTE: This call can take a bit of time
-    return buildjson.query_job_data(complete_at, request_id)
-
-
-def _find_files(job_schedule_info):
+def _find_files(taskId):
```
armenzg, line 271: I want to keep both functions, one for BuildApi and one for TaskCluster. I'm not ready to let go! :smiley: 

Could you please add a note to the old _find_files() saying that it will be deprecated?

In general, I think finding files for a job should live in the respective query_source classes; what do you think?

mozci/sources/tc.py
```diff
@@ -289,3 +290,19 @@ def generate_task_graph(scopes, tasks, metadata):
...
+    result = index.findTask(namespace)
+    return result['taskId']
+
+
+def get_artifacts(taskid, runid=0):
```
armenzg, line 305: docstring please.

test/test_mozci.py
```diff
@@ -16,6 +16,188 @@
...
+    },
+    {
+      "storageType": "s3",
+      "name": "public/build/firefox-43.0a1.en-US.langpack.xpi",
+      "expires": "2016-09-09T07:31:57.620Z",
+      "contentType": "application/x-xpinstall"
+    },
+    {
+      "storageType": "s3",
+      "name": "public/build/firefox-43.0a1.en-US.%(branch)s.checksums",
```
armenzg, line 115: Why are you calling this branch?

I see this:
public/build/firefox-48.0a1.en-US.linux-x86_64.checksums

Are you thinking of "arch"?

test/test_mozci.py
```diff
@@ -16,6 +16,188 @@
...
+    "running_eta": 317,
+    "start_timestamp": 1455883462,
+    "build_architecture": "-",
+    "last_modified": "2016-02-19T12:14:47",
+    "build_platform_id": 144,
+    "job_group_id": 41
+  }]
+"""
+
+MOCK_ARTIFACTS = """
```
armenzg, line 98: From my reading, this is a Buildbot mock artifacts.

We will also need to write tests for the TaskCluster tasks
https://tools.taskcluster.net/task-inspector/#L4N7_qbfSa-0M6Fh4jymCQ/0

Also note that the equivalent platforms from Buildbot and TaskCluster don't use the same naming (e.g. linux64-dbg VS linux64-debug OR linux64-artifact-opt VS linux64-opt). What a mess! I will ask about it.

test/test_mozci.py
```diff
@@ -121,3 +303,56 @@ def test_status_summary_running(self, get_status):
     def test_status_summary_coalesced(self, get_status):
         """Test StatusSummary with a coalesced state."""
         assert mozci.mozci.StatusSummary(self.jobs).coalesced_jobs == 1
+
+
+class TestTriggerfunctions(unittest.TestCase):
```
armenzg, line 308: Could you also add another class to test listing of namespaces?
I would like to test that the firefox, mobile and b2g namespaces are correct.

For instance, if we list the namespaces for "firefox" we should test for the current known list of namespaces. In other words, let's test that we don't have new namespaces. If we get new namespaces it means that a new TaskCluster platform has been added and we should look if we need to make any changes for mozci (e.g. win64-tc-opt build).

[1] https://tools.taskcluster.net/index/#gecko.v2.mozilla-central.latest.firefox/gecko.v2.mozilla-central.latest.firefox

test/test_mozci.py
```diff
@@ -121,3 +303,56 @@ def test_status_summary_running(self, get_status):
...
+
+#    @patch('mozci.mozci.query_repo_name_from_buildername',
+#          return_value="mozilla-inbound")
+#    @patch('mozci.platforms.determine_upstream_builder',
+#           return_value="Ubuntu VM 12.04 x64")
+#    @patch('mozci.mozci.valid_builder', return_value=True)
+#    @patch('mozci.query_jobs.TreeherderApi.get_matching_jobs',
+#           return_value=json.loads(MOCK_JOBS % {'result': "success", 'state': "completed"}))
+#    @patch('mozci.query_jobs.TreeherderApi.get_job_status',
+#           return_value=SUCCESS)
```
armenzg, line 358: Could you please delete this if we don't need it?

mozci/mozci.py
MikeLing, line 168: ok, I will make it happened. :)

MikeLing:
Hi @armenzg , your comments make a lot of sense for me. Thank you for that and I think I need some more time to improve it. Maybe it will looks far more different from this version :)

mozci/sources/tc.py
MikeLing, line 23: oh sorry, I haven't consider about we have other projects. I will address that.

armenzg:
I'm glad it helped.

Let me know when you would like me to do the next review.
Please don't squash the commit so I can see it.

I'm going to start using hub instead of merging PRs:
https://blog.spreedly.com/2014/06/24/merge-pull-request-considered-harmful/#.VwKYD-8tjCL

PR #449, opened 2016-04-02
Add function to invalidate JOBS_CACHE

F3real:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/449)


F3real:
It seems one of tests is failing, but I am not seeing connection to code changes in this PR and failure that is happening.
Edit: Seems that other PRs are failing at the same thing.

MikeLing:
@armenzg see same error on this patch https://github.com/mozilla/mozilla_ci_tools/pull/436

armenzg:
I filed an issue for that test.
The only change that could have introduced the issue would have been this:
https://github.com/mozilla/mozilla_ci_tools/commit/288b525ed5b6f97cf40f004a6837e1c6a2cbdd87

I don't see how it could be related.
Maybe it is a dependency issue or a test that is badly written.
Perhaps for our tox tests we should test twice, once with pinned versions of our python dependencies and another one with the latest versions of our packages.
I will file an issue.

mozci/query_jobs.py
```diff
@@ -110,18 +110,27 @@ def _select_jobs_with_specified_status(
 
 class BuildApi(QueryApi):
 
-    def get_all_jobs(self, repo_name, revision):
+    def get_all_jobs(self, repo_name, revision, useCache=True):
```
armenzg: use_cache
I follow _lower_case_with_underscores_
from https://www.python.org/dev/peps/pep-0008/#naming-conventions

mozci/query_jobs.py
```diff
@@ -110,18 +110,27 @@ def _select_jobs_with_specified_status(
...
         """
         Return a list with all jobs for that revision.
 
         If we can't query about this revision in buildapi_client we return an empty list.
         """
+        if not useCache:
+            JOBS_CACHE[(repo_name, revision)] = \
+                query_jobs_schedule(repo_name, revision, auth=get_credentials())
```
armenzg, line 121: OK. We build the cache but we don't use it (in case others want to use the cache later).

Good idea!

mozci/query_jobs.py
```diff
@@ -110,18 +110,27 @@ def _select_jobs_with_specified_status(
...
         if (repo_name, revision) not in JOBS_CACHE:
             JOBS_CACHE[(repo_name, revision)] = \
                 query_jobs_schedule(repo_name, revision, auth=get_credentials())
 
         return JOBS_CACHE[(repo_name, revision)]
 
+    def invalidate_jobs_cache(self):
+        global JOBS_CACHE
+        JOBS_CACHE = {}
+        return
```
armenzg: No need for a return.

armenzg:
Good job. Very minor changes.

Please make changes, squash and merge without the need for a review. Thanks Stefan!

armenzg:
:smiley:

PR #453, opened 2016-04-07
[WIP]Added second tox enviroment for testing new dependancies

F3real:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/453)


armenzg:
Hey Stefan,
This seems to be doing the job, however, there are a couple of things that I want to do different.

For the current list of dependencies it is better you use this experimental feature of tox [1]
You also need to list every dependency + every dependency of each of the dependencies.
You can see in pulse_actions' requirements.txt that there are direct dependencies and depenencies of dependencies listed [2]
Being said that, could you call it current_requirements.txt and use deps=?

For the case for latest requirements, you don't need to specify the list of requirements inside of tox.in. Simply let it work the way it has always worked.

A negative side effect of listing the dependencies inside of tox.ini is that if the list of requirements in setup.py changed we would also have to keep tox.ini up-to-date.

Let me know if you have any objections.

[1] http://tox.readthedocs.org/en/latest/example/basic.html#depending-on-requirements-txt
[2] https://github.com/mozilla/pulse_actions/blob/master/requirements.txt

---

Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/453#-:-KEktfq_-pXvmmdEeQvA:785745851)_

F3real:
@armenzg 
Could you take a look at the changes made ?
I made py27_new_deps don't run automatically. But they still can be run with:
**tox -e py27_new_deps**

setup.py
```diff
@@ -1,5 +1,14 @@
 from setuptools import setup, find_packages
 
+#When tox is run, relative directory is changed and IOError is thown when trying to open requirements.
+#We can safely ignore this due the fact that tox calls install with --no-deps flag.
+required = []
+try:
+    with open('requirements.txt') as f:
+        required = f.read().splitlines()
+except IOError:
```
armenzg: We want to fail if we were unable to read the requirements.txt

requirements.txt
```diff
@@ -0,0 +1,9 @@
+buildapi_client==0.2.2
```
armenzg: If requirements.txt is the _official_ list of depencies then we should not fix the versions.

tox.ini
```diff
@@ -1,12 +1,14 @@
...
-envlist = py27
+envlist = py27_current_deps
 
 [testenv]
 basepython = python
 
+[testenv:py27_current_deps]
+basepython = python2.7
 deps =
+    -rrequirements.txt
```
armenzg: Could we add a second file current_requirements.txt with fixed versions?

tox.ini
```diff
@@ -16,6 +18,29 @@ commands =
     flake8 mozci test
     coverage run --source=mozci -m py.test test
 
+
+[testenv:py27_new_deps]
+basepython = python2.7
+deps =
```
armenzg, line 34: In here we could use requirements.txt which would not have fixed versions.

armenzg:
This is looking good.

Could you also make Travis to run _both_ configurations? Thanks!

setup.py
F3real: @armenzg 
Problem is that tox invocation fails in that case. Tox is unable to locate requirements.txt.

armenzg: OK let me know when you want me to look at this patch again. Thank you Stefan!

RonnyPfannschmidt:
this package is **broken** until you add a inclusion of `requirements.txt` to the `MANIFEST.in`

also its best-practice to have pinned versions in the requirements and unpinned versions in `setup.py`

armenzg:
Thanks @RonnyPfannschmidt !

Let's go back to having the dependencies listed in setup.py
Let's also have the "current" dependencies directly in tox's env.

In the future, when we need to test newer releases we will just update the tox env.
If new packages need to be added to mozci we will have to add in both places (setup.py and tox's env).

@F3real does this make sense to you?

armenzg:
If we don't use requirements.txt we won't need to touch Makefile.in

armenzg:
Reviewed 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_[tox.ini, line 41 [r4]](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/453#-KFt6ibXwN9XE42CYxNl:-KFt6ibYGe3cbqCdUBlR:1444836434) ([raw file](https://github.com/mozilla/mozilla_ci_tools/blob/cc50912c17794c1e2d8f39ac9b1db4ae30f3e645/tox.ini#L41)):_
I don't think you need to list the dependencies in here since it will use whatever is declared on setup.py.
In the env above, we use it because we need to pin the versions rather than letting them get installed via setup.py

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/453)_

armenzg:
We only need to remove those deps of the second env to finish up this PR. Thanks Stefan!

---

Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

_Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/453#-:-KFt79EiDHTU-ra5uBw7:-679836182)_

armenzg:
We might also want to consider adding a [base] section for packages shared by both envs:
https://hg.mozilla.org/mozilla-central/file/default/testing/mozharness/tox.ini#l4

PR #482, opened 2016-06-03
Adding extra property before triggering job for nightly build

martiansideofthemoon:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/482)

### DIFF
--- a/mozci/sources/buildbot_bridge.py
+++ b/mozci/sources/buildbot_bridge.py
@@ -359,7 +359,7 @@ def _generate_tc_tasks_from_builders(builders, repo_name, revision):
             metadata = get_buildername_metadata(builder)
             if metadata['platform_name'].startswith('android') and \
                metadata['nightly'] is True and \
-               'l10n' in builder:
+               'l10n' not in builder:
                 properties = {}
             task = _create_task(
                 buildername=builder,
@@ -549,7 +549,7 @@ def _generate_tasks(repo_name, revision, builders_graph, metadata=None, task_gra
         metadata = get_buildername_metadata(builder)
         if metadata['platform_name'].startswith('android') and \
            metadata['nightly'] is True and \
-           'l10n' in builder:
+           'l10n' not in builder:
             upload_to_task_id = {}
         task = _create_task(
             buildername=builder,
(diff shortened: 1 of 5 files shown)


martiansideofthemoon:
**r?** @armenzg ?

mozci/sources/buildbot_bridge.py
```diff
@@ -354,12 +354,19 @@ def _generate_tc_tasks_from_builders(builders, repo_name, revision):
     # builders already on our list
     for builder in builders:
         if is_upstream(builder):
+            properties = {'upload_to_task_id': slugId()}
+            # Bug 1274483 adds this condition
+            metadata = get_buildername_metadata(builder)
+            if metadata['platform_name'].startswith('android') and \
+               metadata['nightly'] is True and \
+               'l10n' in builder:
```
armenzg: There's no l10n in the builder name,

The nightly build does both en-US and multi-locale versions. Even though there will be L10n builders triggered by it.

mozci/sources/buildbot_bridge.py
```diff
@@ -537,6 +544,13 @@ def _generate_tasks(repo_name, revision, builders_graph, metadata=None, task_gra
         # Due to bug 1221091 this will be used to know to which task
         # the artifacts will be uploaded to
         upload_to_task_id = slugId()
+
+        # Bug 1274483 adds this condition
+        metadata = get_buildername_metadata(builder)
+        if metadata['platform_name'].startswith('android') and \
+           metadata['nightly'] is True and \
+           'l10n' in builder:
```
armenzg: No need for 'l10n'

setup.py
```diff
@@ -7,7 +7,7 @@
     "keyring",
     "progressbar",
     "pushlog_client",
-    "requests",
+    "requests[security]",
```
armenzg: We don't need the optional security module for this PR.

The warnings I would get went away when I upgraded from Ubuntu 12.04 to 16.04 since the /usr/bin/python version was higher.

armenzg:
Good change. Address my comments and this should be ready to be merged.

Thanks!

setup.py
martiansideofthemoon: I'm stuck on a 14.04

PR #483, opened 2016-06-05
Bug 1276914 - Added new handler for missing builder

martiansideofthemoon:
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/483)


armenzg:
What's the reasoning behind changing the name of the error?

It is indeed more clear but is there anything else beyond that?

PR #493, opened 2016-07-12
Switch to using TreeherderClient's new server_url argument

edmorley:
treeherder-client 3.0.0 has switched from using separate `host` and `protocol` arguments to a single `server_url` one (https://github.com/mozilla/treeherder/commit/6001db37614f8b41c97499d74b3aac9f749e0bff). As before, this argument can be omitted to default to production Treeherder.

Whilst treeherder-client 3.0.0 supports the old style arguments, they're deprecated.

In `TreeherderApi` I've stuck with the same argument name as used in `TreeherderClient` for greater consistency. In addition, calling `TreeherderApi(treeherder_server_url='foo')` feels a little redundant, since the class is only for Treeherder, so it's already clear about what `server_url` is referring.

---

This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/493)


armenzg:
Thank you sir!

edmorley:
You're welcome :-)

edmorley:
Meant to say - treeherder-client 3.0.0 also now uses connection pooling, so if consumers of mozci's re-use the same `TreeherderApi` instance for multiple requests in a row, they'll get a performance boost from HTTP keep-alive.

armenzg:
Good news all the way! 💃

PR #504, opened 2016-11-28
Bug 1289823 - Flexible action task

imbstack:
@armenzg: This should be updated here and pulled into pulse_actions before https://reviewboard.mozilla.org/r/96072 lands. It will allow it to work with both styles of action.yml until the old-style is completely gone at some point. I have backfill logic in-tree ready to go once this is all merged in if this is a good direction to go in.

Heads up: @djmitche

The added lines in `test/helpers.py` was just to get `tox` to pass locally. I assume some lint library was updated at some point or something.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/504)

### DIFF
--- a/mozci/taskcluster/tc.py
+++ b/mozci/taskcluster/tc.py
@@ -114,7 +114,7 @@ class TaskClusterManager(BaseCIManager):
 
         # Satisfying mustache template variables in YML file
         # We must account for both the old and new style of arg passing
-        if action_args:
+        if action_args and "{{action_args}}" in action_task:
             action_args = " ".join(["--{}='{}'".format(k, v) for k, v in action_args.items()])
             action_task = action_task.replace("{{action_args}}", action_args)
         else:
--- a/mozci/taskcluster/tc.py
+++ b/mozci/taskcluster/tc.py
@@ -102,7 +102,7 @@ class TaskClusterManager(BaseCIManager):
         else:
             LOG.info("We did not schedule anything because we're running on dry run mode.")
 
-    def schedule_action_task(self, decision_task_id=None, task_labels=None, action_args=None):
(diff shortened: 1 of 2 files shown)


armenzg:
This lgtm. Just address the question wrt the optional parameter and we can merge and land after that.
I can take care of bumping the mozci version and we need pulse_actions to use this new version.

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[mozci/taskcluster/tc.py, line 112 at r1](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/504#-KXkRQiAL2kXpSHVFJ4P:-KXkRQiBwAZsehI4lYe0:b-3s7ktm) ([raw file](https://github.com/mozilla/mozilla_ci_tools/blob/2663fcd12de7b31292d1f6b63a5352afab79b353/mozci/taskcluster/tc.py#L112)):*
> ```Python
>         # Downloading the YAML file for action tasks
>         # Action Tasks will be used to schedule the Task Labels in the parameters
>         action_task = get_artifact_for_task_id(task_id=decision_task_id,
> ```

We've changed the signature to make decision_task_id optional, however, in this line we still require it at all times.
Could you please keep decision_task_id as mandatory? Unless I'm missing something.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/mozilla/mozilla_ci_tools/504#-:-KXkTAnpQWSVwvCcgu9Z:bgd1fbs)*

djmitche:
Will one of the action_args be "action" or some such thing that says which action to perform?

imbstack:
@djmitche: Yep, that's the plan!

PR #505, opened 2016-12-08
Improve action-task arguments

imbstack:
requested reviewer: @armenzg 

@djmitche and I talked about how we'd like the arguments to the action task to look a bit today and settled on something along the lines of `mach taskgraph {{action}} {{action_args}}` which would end up being something like `mach taskgraph backfill --foo=bar`. This is a change to support that style. I'll have a PR for pulse_actions shortly after this is published.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/505)

### DIFF
--- a/mozci/taskcluster/tc.py
+++ b/mozci/taskcluster/tc.py
@@ -102,7 +102,7 @@ class TaskClusterManager(BaseCIManager):
         else:
             LOG.info("We did not schedule anything because we're running on dry run mode.")
 
-    def schedule_action_task(self, decision_task_id, task_labels=None, action_args=None):
+    def schedule_action_task(self, action, decision_task_id, action_args=None):
         """
         Function which will be used to schedule an action task.
         Action Tasks use in-tree logic to schedule the task_labels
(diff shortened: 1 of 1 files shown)


imbstack:
Not sure why the tests are failing, but it appears to have something to do with downloading `allthethings.json`. I don't think it is related to my changes.

armenzg:
You're right.
The file has been removed from releng's systems.
I will have to point to a newer one.

mozci/taskcluster/tc.py
```diff
@@ -114,12 +114,13 @@ def schedule_action_task(self, decision_task_id, task_labels=None, action_args=N
...
         # We must account for both the old and new style of arg passing
-        if action_args and "{{action_args}}" in action_task:
+        if "{{action_args}}" in action_task:
             action_args = " ".join(["--{}='{}'".format(k, v) for k, v in action_args.items()])
+            action_task = action_task.replace("{{action}}", action)
             action_task = action_task.replace("{{action_args}}", action_args)
         else:
             action_task = action_task.replace("{{decision_task_id}}", decision_task_id)
-            action_task = action_task.replace("{{task_labels}}", ",".join(task_labels))
+            action_task = action_task.replace("{{task_labels}}", ",".join(action_args["task_labels"]))
```
armenzg, line 123: We can change [this](https://github.com/mozilla/pulse_actions/blob/1348146fc10dc48efbb6ed282f0e0cfcba00ce33/pulse_actions/handlers/treeherder_add_new_jobs.py#L120) once this is released.

armenzg:
I released this as 0.52.0

PR #506, opened 2016-12-14
Fix single-quotes in yaml template

imbstack:
@armenzg: sorry, this needs changed one more time. The single quotes wrapping the variables were getting inserted into something that was already getting rendered with single quotes around it so it was breaking yaml rendering. I've added tests for this section now though, so we should not have to change this again!

The change of `decision_task_id` to `decision_id` was made to match this to the in-tree `action.py` more closely without requiring us to do special-casing in this code.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/506)


armenzg:
I've released a new version, bumped pulse_actions and deployed it to production :)

PR #513, opened 2017-09-09
bump the tc version number to 4.0.0 on tox

MikeLing:
Hi @armenzg , 

So after I set up my env by `python setup.py ` and `run tox`, I got this error https://pastebin.mozilla.org/9031957. It seems like the tox require treeherder-client 2.0.1 but the latest number is 4.0.0. I'm not sure if it's a bug, but after I change the version number to 4.0.0, everything works well.

Please close this pr if I misunderstand something.  Thank you.

---
This change is [Reviewable](https://reviewable.io/reviews/mozilla/mozilla_ci_tools/513)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -25,7 +25,7 @@ deps =
     PyYAML==3.11
     requests==2.8.1
     taskcluster==0.3.0
-    treeherder-client==2.0.1
+    treeherder-client==4.0.0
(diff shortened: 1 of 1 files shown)


armenzg:
Thanks MikeLing but I will be changing the README saying we're not developing it further.

